PR #204, opened 2014-04-04
Add a logs modal to UI

mrdmnd:


redox:
Hey @mrdmnd, any chance you could share the way you're using Chronos to read the logs of the underlying Mesos?

I saw the [add_chronos_logs](https://github.com/mesos/chronos/tree/add_chronos_logs) branch, but it doesn't seem to include the code to fetch the logs, only part of the display. Some guys are discussing about it here as well: http://stackoverflow.com/questions/28669547/how-to-generate-the-url-which-shows-the-live-logs-in-mesos-for-a-job

Thanks!

mrdmnd:
Hey, @redox - I no longer am a maintainer on Chronos after switching to a new company, but I'd be willing to dig back through some old code and see if I can figure it out. Give me a day or so?

redox:
That would be cool, thank you for your help :)

PR #319, opened 2014-12-11
Allow launching multiple tasks per resource offer

meelapshah:
Mesos allows frameworks to respond to offers with a list of tasks to run.

Currently, Chronos launches at most task per offer. I think it would be helpful to respond with many tasks from the queue per offer to reduce the delay between getting added to the queue and being launched on Mesos.

This patch accomplishes this in a simple way by repeatedly popping tasks from the queue as long as there are resources remaining to satisfy them, and then launches them all together.


brndnmtthws:
Okay, looks good.  Thanks for the patch.

PR #326, opened 2014-12-24
Add leader's url to proxy log

depay:
Add leader's url to proxy log to to check whether proxy to the right one.

### DIFF
--- a/src/main/scala/com/airbnb/scheduler/jobs/JobScheduler.scala
+++ b/src/main/scala/com/airbnb/scheduler/jobs/JobScheduler.scala
@@ -675,12 +675,19 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
 
     running.set(true)
     lock.synchronized {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/airbnb/scheduler/jobs/JobScheduler.scala
```diff
@@ -675,12 +675,19 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
 
     running.set(true)
     lock.synchronized {
-      //It's important to load the tasks first, otherwise a job that's due will trigger a task right away.
-      log.info("Loading tasks")
-      TaskUtils.loadTasks(taskManager, persistenceStore)
-
-      log.info("Loading jobs")
-      JobUtils.loadJobs(this, persistenceStore)
+      try {
```
brndnmtthws: Could you please open a separate PR for this change?

PR #337, opened 2015-01-10
Support for custom command arguments when manually running a Chronos job

doronin:
This commit adds support for passing custom command arguments when manually running a Chronos job using `PUT /scheduler/<jobName>` endpoint

see https://github.com/airbnb/chronos/issues/335

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ chronos.arx
 chronos.tbz
 .classpath.txt
 dependency-reduced-pom.xml
+.arcconfig
--- a/pom.xml
+++ b/pom.xml
@@ -380,7 +380,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>${maven-surefire-plugins.version}</version>
         <configuration>
-          <argLine>-Xmx1024m</argLine>
+          <argLine>-Xmx1024m -Duser.timezone=UTC</argLine>
           <redirectTestOutputToFile>false</redirectTestOutputToFile>
           <includes>
             <include>**/Test*.java</include>
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ You can manually start a job by issuing an HTTP request.
 
 * Endpoint: __/scheduler/job__
 * Method: __PUT__
-* Query string parameters: `arguments` - optional string with a list of command line arguments that is appended to job's `command` (if `shell` is true) or `arguments` (if `shell` is false)
+* Query string parameters: `arguments` - optional string with a list of command line arguments that is appended to job's `command`
 * Example: `curl -L -X PUT chronos-node:8080/scheduler/job/request_event_counter_hourly`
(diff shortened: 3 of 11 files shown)


pom.xml
```diff
@@ -380,7 +380,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>${maven-surefire-plugins.version}</version>
         <configuration>
-          <argLine>-Xmx1024m</argLine>
+          <argLine>-Xmx1024m -Duser.timezone=UTC</argLine>
```
brndnmtthws: Thanks for that!

.gitignore
```diff
@@ -13,3 +13,4 @@ chronos.arx
 chronos.tbz
 .classpath.txt
 dependency-reduced-pom.xml
+.arcconfig
```
brndnmtthws: No newline at EOF.

brndnmtthws:
This looks okay, but you'll also need to update the documentation to reflect the change.

doronin:
Brenden thanks for the review. I'll change docs and update this diff

doronin:
Updated docs (https://github.com/doronin/chronos/tree/args#manually-starting-a-job) and added new line character at the end of .gitignore

brndnmtthws:
Just out of curiosity, what is the rationale for this?

You could accomplish the same thing simply by changing the command as needed.

doronin:
The rationale is laziness :) I needed a way to run a script with custom arguments on-demand. Ideally this functionality could be implemented in a separate framework but:
* adding this little modification was faster as Chronos already has a way to start a job manually
* Chronos is already deployed so there's no need to add extra piece of infrastructure
* Chronos is well tested

brndnmtthws:
Okay, seems reasonable

Any objections from @airbnb/di or @elingg?

elingg:
It does seem reasonable, but I think it's better to keep as part of the command to avoid strange errors. In this case, since we run both jobs that are commands and also docker containers, the args can depend on the type of command (docker or otherwise) and on the entry point. Args can be evaluated differently so to me, the safest way is to keep them as part of the command as opposed to adding a new variable. Feel free to discuss!

doronin:
@elingg I can change [getJobWithArguments] (https://github.com/doronin/chronos/blob/args/src/main/scala/com/airbnb/scheduler/jobs/JobUtils.scala#L161) method to always append custom `arguments` to `command` parameter regardless if `job.shell` is true or false

elingg:
@doronin if you append the automatically append the arguments to the command in all cases, we would avoid unexpected behavior. That sounds better to me, thx!

doronin:
@elingg I made the change you requested and updated README to reflect the change

brndnmtthws:
Thanks folks.

elingg:
thank you @doronin! in the readme would you mind fixing the following example to make debug%20true debug=true. It looks a little strange.

Example: curl -L -X PUT chronos-node:8080/scheduler/job/job_name?arguments=-debug%20true

PR #347, opened 2015-01-22
Disable flakey test

BenWhitehead:


BenWhitehead:
@elingg @brndnmtthws Can either of you merge this PR?

This flakey test is blocking @lingmann getting new packages out.

brndnmtthws:
Do you have a log of the failure?

BenWhitehead:
Here is a link to all the build failures jeremy ran into.

https://teamcity.mesosphere.io/viewType.html?buildTypeId=Oss_Chronos_MakeDistroPackage_2&tab=buildTypeHistoryList&branch_Oss_Chronos=__all_branches__

brndnmtthws:
I'd like to figure out *why* it failed, and after investigating for a bit, it's not clear what the cause is. I'm not opposed to merging this if that's what's needed to move things forward.

BenWhitehead:
True I think this should be tracked down, however the workaround that was in place before was to disable all tests.  It's better to have one disabled test then not run any at all.  I don't know what would be involved in tracking down and fixing this test, I opened the PR to help remove the blocker for @lingmann.

elingg:
@benwhitehead, give me a few to investigate

brndnmtthws:
I've been running this test in a loop on my laptop for about an hour now, without any errors. Something is fishy.

elingg:
Same here, I can't recreate the problem on my laptop. It could be a difference in some state configuration on the teamcity environment, possibly related to zookeeper. It seems maybe the test isn't well written, but I would like to track down why.

elingg:
Debugged the issue. The test was using a ScheduledThreadPool as a listener which was not working in all build environments.. I will be creating a PR to fix the test so I will close this PR. Thanks!

PR #350, opened 2015-01-26
Increase default failover timeout to 1 week

elingg:
This will match Marathon’s default.


brndnmtthws:
Seems reasonable.

:+1:

PR #351, opened 2015-01-27
Reorganize packages, reformat code.

brndnmtthws:
This is a code reorg to reflect that the project is no longer owned by Airbnb.

cc @elingg


brndnmtthws:
Okay, I can do that instead. I went with github because the project is hosted on github.

brndnmtthws:
Actually, `org.mesos` doesn't really make sense since mesos.org doesn't exist. I'd rather just leave it like this.

brndnmtthws:
Okay, I changed it to `org.apache.mesos.chronos`, which is a bit verbose for my liking, but it should suffice.

elingg:
Thank you! :+1:

PR #352, opened 2015-01-27
Job retry policy

brndnmtthws:
This depends on #351.

cc @elingg


brndnmtthws:
I don't think there should be any other synchronization issues, since this is the only place where a method is called from another place (and I've synchronized that particular case).

elingg:
Ok, it looks like we are good. The only other action on the queue called externally is the refresh which is called from a synchronized method. 

 def reset(purgeQueue: Boolean = false) {
    lock.synchronized {
      streams = List()
      jobGraph.reset()
      if (purgeQueue) {
        log.warning("Purging locally queued tasks!")
        taskManager.flush()
      }
    }
  }

elingg:
:+1: thanks for fixing this!

PR #353, opened 2015-01-29
Add basic auth support for http DELETE command

stieg:
Forgot to add this when I added support to the PUT command.


brndnmtthws:
Thanks!

PR #354, opened 2015-01-30
Fixing the package to execute the main scheduler.

chikin:
The Readme still had the old package structure for the main class.


elingg:
LGTM! thanks for your fix

PR #356, opened 2015-01-30
There were more instances of the old package structure

chikin:
The code had more instances of using the old package structure com.airbnb instead of the current org.apache.mesos package.


elingg:
thanks for catching this! I will do a quick check to see if anything else remains. :shipit

Issue #355, opened 2015-01-30
null field appended to command


yoeduardoj:
We submit a very simple job, and when it gets executed "null" gets appended to the command.

```json
{
  "async": false,
  "command": "echo \"Hello Mini\"",
  "cpus": 1,
  "disk": 4096,
  "epsilon": "PT30M",
  "executor": "",
  "mem": 5000,
  "name": "MiniCurl",
  "owner": "a@gmail.com",
  "retries": 0,
  "runAsUser": "mesos",
  "schedule": "R/2014-11-07T18:40:42Z/PT24H"
}
```

The task ID created in Mesos already looks funny: ct:1422586155844:0:MiniCurl:null	

The command succeeds in this case, but looking closely at how it was executed:

$ cat stdout
Registered executor on ip-10-250-100-157.ec2.internal
Starting task ct:1422582459084:0:MiniCurl:null
Forked command at 10385
sh -c 'echo "Hello Mini" null'
Hello Mini null
Command exited with status 0 (pid: 10385)
$ cat stderr
I0130 01:47:40.187083 10363 exec.cpp:132] Version: 0.21.1
I0130 01:47:40.189077 10378 exec.cpp:206] Executor registered on slave 20150129-184612-1080359434-5050-1729-S0
$

Notice:
sh -c 'echo "Hello Mini" null'

Where did the null come from? Tried the same command from the UI works with no null appended, and the task Id looks better

"ct:1422587418000:0:MiniCurl2:	"

I looked at how the UI posts to the chronos API, and here's the payload:

```json
{
  "schedule": "R/2015-01-30T03:10:18Z/PT1M",
  "epsilon": "PT30M",
  "name": "MiniCurl2",
  "command": "echo \"Hello Mini\"",
  "owner": "a@gmail.com",
  "async": false,
  "executor": "",
  "disabled": false,
  "softError": false
}
```

laktech:
I see the same thing occurring when suing chrono-sync.rb with a yaml configuration. Note also the "null" in the task name.

```
Starting task ct:1422658619492:0:Hello:null
Forked command at 8043
sh -c 'echo Hello null'
Hello null
Command exited with status 0 (pid: 8043)
```

```
---
name: Hello
command: echo Hello
shell: true
owner: a@foo.com
cpus: 0.05
disk: 1024.0
mem: 720.0
softError: false
uris: []
environmentVariables: []
highPriority: false
runAsUser: root
[...]

PR #360, opened 2015-02-02
Fix null args, variable shadowing.

brndnmtthws:
cc @elingg

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
@@ -90,8 +90,8 @@ object TaskUtils {
     })
   }
 
-  def getTaskId(job: BaseJob, due: DateTime, attempt: Int = 0, arguments: String = ""): String = {
-    taskIdTemplate.format(due.getMillis, attempt, job.name, arguments)
+  def getTaskId(job: BaseJob, due: DateTime, attempt: Int = 0): String = {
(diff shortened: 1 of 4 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
```diff
@@ -90,8 +90,8 @@ object TaskUtils {
     })
   }
 
-  def getTaskId(job: BaseJob, due: DateTime, attempt: Int = 0, arguments: String = ""): String = {
-    taskIdTemplate.format(due.getMillis, attempt, job.name, arguments)
+  def getTaskId(job: BaseJob, due: DateTime, attempt: Int = 0): String = {
+    taskIdTemplate.format(due.getMillis, attempt, job.name, job.arguments.foldRight("")((a, b) => if (b.nonEmpty) s"$a $b" else a))
   }
```
elingg: This makes sense, but could we simplify the logic some by iterating through the list from left to right and doing a simple string concatenation. job.arguments.foldLeft("") ((a, b) => if (b.nonEmpty) a + " " + b else a))

elingg:
Thank you for fixing this. This should fix https://github.com/mesos/chronos/issues/355. I have only one comment above.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
brndnmtthws: You're probably right. I got a bit carried away.

tamizhgeek:
Thanks guys!

PR #361, opened 2015-02-02
Scala style cleanup.

brndnmtthws:
This includes a whole bunch of Scala style fixes, including replacing
deprecated code with "the right way".

cc @elingg


brndnmtthws:
Hey @elingg, can you take a look at this when you get a chance?

elingg:
@brndnmtthws, yes, will finish reviewing it soon

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
```diff
@@ -41,8 +42,7 @@ class TaskManager @Inject()(val listeningExecutor: ListeningScheduledExecutorSer
 
   val taskCache = CacheBuilder.newBuilder().maximumSize(5000L).build[String, TaskState]()
 
-  val taskMapping = new mutable.HashMap[String, mutable.ListBuffer[(String, Future[_])]] with
-    collection.mutable.SynchronizedMap[String, mutable.ListBuffer[(String, Future[_])]]
+  val taskMapping: concurrent.Map[String, mutable.ListBuffer[(String, Future[_])]] = new ConcurrentHashMap().asScala
```
elingg: I agree that this should be updated to be a concurrent hash map, but this probably means we can get rid of some of the synchronized blocks in this class. Perhaps this should be a separate PR.

src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala
```diff
@@ -18,9 +21,9 @@ import scala.collection.mutable.{HashMap, ListBuffer, Map, SynchronizedMap}
 @ThreadSafe
 class JobGraph {
   val dag = new DirectedAcyclicGraph[String, DefaultEdge](classOf[DefaultEdge])
-  val edgeInvocationCount = Map[DefaultEdge, Long]()
+  val edgeInvocationCount = mutable.Map[DefaultEdge, Long]()
   private[this] val log = Logger.getLogger(getClass.getName)
-  private[this] val jobNameMapping = new HashMap[String, BaseJob] with SynchronizedMap[String, BaseJob]
+  private[this] val jobNameMapping: concurrent.Map[String, BaseJob] = new ConcurrentHashMap().asScala
```
elingg: I agree that this should be a synchronized hash map, but this means that we can get rid of some of the synchronized blocks in the code. Perhaps this should be a separate PR.

src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala
```diff
@@ -18,9 +21,9 @@ import scala.collection.mutable.{HashMap, ListBuffer, Map, SynchronizedMap}
 @ThreadSafe
 class JobGraph {
   val dag = new DirectedAcyclicGraph[String, DefaultEdge](classOf[DefaultEdge])
-  val edgeInvocationCount = Map[DefaultEdge, Long]()
+  val edgeInvocationCount = mutable.Map[DefaultEdge, Long]()
```
elingg: Why does this map need to be mutable?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/graph/Exporter.scala
```diff
@@ -16,7 +17,7 @@ object Exporter {
 
   def export(w: Writer, jobGraph: JobGraph) {
     val dag = jobGraph.dag
-    val jobMap = new HashMap[String, BaseJob]
+    val jobMap = new mutable.HashMap[String, BaseJob]
```
elingg: Same question- why does this need to be mutable?

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala
```diff
@@ -31,7 +32,7 @@ class MesosJobFramework @Inject()(
   val frameworkName = "chronos"
   private[this] val log = Logger.getLogger(getClass.getName)
   private var lastReconciliation = DateTime.now.plusSeconds(config.reconciliationInterval())
-  private var runningTasks = new HashMap[String, ChronosTask]
+  private var runningTasks = new mutable.HashMap[String, ChronosTask]
```
elingg: Again same question with the amps in this class. What is the design choice for making them mutable?

src/test/scala/org/apache/mesos/chronos/scheduler/jobs/ScheduleStreamSpec.scala
```diff
@@ -12,17 +12,17 @@ class ScheduleStreamSpec extends SpecificationWithJUnit {
       val orgSchedule = "R3/2012-01-01T00:00:00.000Z/P1D"
       val stream = new ScheduleStream(orgSchedule, null)
       stream.head must_==(orgSchedule, null, "")
-      stream.tail.get.head must_==("R2/2012-01-02T00:00:00.000Z/P1D", null, "")
-      stream.tail.get.tail.get.head must_==("R1/2012-01-03T00:00:00.000Z/P1D", null, "")
-      stream.tail.get.tail.get.tail.get.head must_==("R0/2012-01-04T00:00:00.000Z/P1D", null, "")
-      stream.tail.get.tail.get.tail.get.tail must_== None
+      stream.tail().get.head must_==("R2/2012-01-02T00:00:00.000Z/P1D", null, "")
```
elingg: Why do we need the parans here?

elingg:
Thank you for cleaning up the style!

brndnmtthws:
For all of the mutable changes, there is no actual code change. It's just a style change (i.e., specifying `mutable.` rather than using an implicit import).

elingg:
Ok about `mutable`, sounds fair then.

src/test/scala/org/apache/mesos/chronos/scheduler/jobs/ScheduleStreamSpec.scala
brndnmtthws: There's a note here on why we do this: http://docs.scala-lang.org/style/method-invocation.html

brndnmtthws:
I'll open a separate PR for the removal of the deprecated `SynchronizedMap` usage.

src/test/scala/org/apache/mesos/chronos/scheduler/jobs/ScheduleStreamSpec.scala
elingg: I see. It is indeed the appropriate style:

"Scala allows the omission of parentheses on methods of arity-0 (no arguments). However, this syntax should only be used when the method in question has no side-effects (purely-functional). In other words, it would be acceptable to omit parentheses when calling queue.size, but not when calling println(). This convention mirrors the method declaration convention given above."

elingg:
Sounds good, once you have separated out the SynchronizedMap usage, feel free to merge. :+1:

brndnmtthws:
Updated PR sans `SynchronizedMap` change.

PR #363, opened 2015-02-03
Add leader's url to proxy log

depay:
Add leader's url to proxy log to to check whether proxy to the right
one.


brndnmtthws:

PR #364, opened 2015-02-03
Fix: Add try-catch to loading tasks and jobs

depay:
Chronos will be unable to decline offers from mesos and will hold all
the mesos resources if exception happends on loading tasks or jobs.
It may be caused by bad network to ZKServer or some other reasons,
exit with code 1.


brndnmtthws:
Looks good! Any comments, @elingg?

elingg:
It also looks good to me. It seems fair that the framework scheduler would exit when we experience this failure.

PR #365, opened 2015-02-03
Fix: cleaning invalid task with a wrong task name

depay:
Func removeTask uses task name without prefix "T_".


brndnmtthws:
Looks reasonable. Have you tested this?

depay:
@brndnmtthws  Tested.

elingg:
@depay, Could you put val taskId = f.substring(taskPrefix.size) above the if statement so taskId can be used here to make the code cleaner?

elingg:
:+1:

PR #366, opened 2015-02-03
Remove invalid tasks on scheduling

depay:
Task will be loaded on restarting. It's meaningless.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
@@ -83,9 +83,11 @@ class TaskManager @Inject()(val listeningExecutor: ListeningScheduledExecutorSer
       }
 
       //If the job was deleted after the taskId was added to the queue, the task could be empty.
-      if (jobOption.isEmpty || jobOption.get.disabled) {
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Looks good. Tested?

depay:
@brndnmtthws  Tested. BTW, if jobOption.get.disabled returns None, how to trigger a disabled job ?

elingg:
@depay, another question about this patch. What if the user reenables a disabled job? Will it be triggered?

depay:
@elingg , i don't think we should care about that. If a task was in queue before and removed, it is out of due more or less at current time.

elingg:
@depay, fair enough when it gets enabled, it will be added to the queue.

depay:
@elingg , in the current code, a task of disabled job is popped and never reenqueued, util loading task during restart . And i'll submit another commit to handle it.

elingg:
@depay thanks for clarifying. I wanted to make sure that when enabling the job, it gets added to the queue. That or we keep it in the queue. Could you please make sure the build succeeds and also retest? Thx!

depay:
@elingg , sorry for my fault with braces and brackets in second commit -_-#

elingg:
LGTM

PR #368, opened 2015-02-03
Remove deprecated synchronizedmap

brndnmtthws:
cc @elingg

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala
@@ -1,6 +1,7 @@
 package org.apache.mesos.chronos.scheduler.graph
 
 import java.io.StringWriter
+import java.util.concurrent.ConcurrentHashMap
 import java.util.logging.Logger
 import javax.annotation.concurrent.ThreadSafe
 
@@ -9,7 +10,8 @@ import org.jgrapht.experimental.dag.DirectedAcyclicGraph
 import org.jgrapht.ext.{DOTExporter, IntegerNameProvider, StringNameProvider}
 import org.jgrapht.graph.DefaultEdge
 
-import scala.collection._
+import scala.collection.convert.decorateAsScala._
+import scala.collection.{mutable, _}
 import scala.collection.mutable.ListBuffer
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala
```diff
@@ -18,9 +21,9 @@ import scala.collection.mutable.{HashMap, ListBuffer, Map, SynchronizedMap}
 @ThreadSafe
 class JobGraph {
   val dag = new DirectedAcyclicGraph[String, DefaultEdge](classOf[DefaultEdge])
-  val edgeInvocationCount = Map[DefaultEdge, Long]()
+  val edgeInvocationCount = mutable.Map[DefaultEdge, Long]()
   private[this] val log = Logger.getLogger(getClass.getName)
-  private[this] val jobNameMapping = new HashMap[String, BaseJob] with SynchronizedMap[String, BaseJob]
+  private[this] val jobNameMapping: concurrent.Map[String, BaseJob] = new ConcurrentHashMap().asScala
```
elingg: Do we still need the synchronized blocks in this class around jobNameMapping since it is a ConcurrentHashMap now?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
```diff
@@ -41,8 +42,7 @@ class TaskManager @Inject()(val listeningExecutor: ListeningScheduledExecutorSer
 
   val taskCache = CacheBuilder.newBuilder().maximumSize(5000L).build[String, TaskState]()
 
-  val taskMapping = new mutable.HashMap[String, mutable.ListBuffer[(String, Future[_])]] with
-    collection.mutable.SynchronizedMap[String, mutable.ListBuffer[(String, Future[_])]]
+  val taskMapping: concurrent.Map[String, mutable.ListBuffer[(String, Future[_])]] = new ConcurrentHashMap().asScala
```
elingg: similar question: Do we still need the synchronized blocks in this class around taskMapping since it is a ConcurrentHashMap now?

brndnmtthws: In this case it's required for `enqueue`.

src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala
brndnmtthws: Nope, I don't believe so. I'll drop it.

brndnmtthws:
@elingg Updated as per comments, PTAL.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
elingg: This is true in enqueue, because the synchronization is on queues and not on taskMapping. Thx for double checking this.

elingg:
LGTM :shipit:

Issue #367, opened 2015-02-03
Job Parameters


jfenc91:
It would be great if Chronos supported task flow level parameters that are accessible from all tasks within a chain of dependent jobs. While the job arguments that were recently added go some way to address this, they have several shortcomings:
- They do not get passed to dependent tasks. 
- There is a limit of around 255 characters because they are passed through the task id which is used as a directory name. 

I propose that we add something to allow the following:
Given a job with arguments such as: “arguments”:[“--p1 ${val1}”, “--p2 ${val2}”]
Request a job to run with flow parameters such as: p1=arg1, p2=arg2 . The job would then execute with parameters: “--p1 arg1 --p2 arg2”.
   
It looks like this could be done by:
- Changing JobManagmentResource’s trigger to a post request and allow it to accept a set of environment variables 
- Storing the state for each task flow in zk with the set of requested environment variables
- Passing along the zk node id for the task flow via the task name 
- Adding the environment variables to the mesos task builder

Some things that may be nice to have with this are:
- Now that each task is associated with its own state, it seems we can run more than one instance of a job at a time. (This will likely require additional changes in the job scheduler to properly handle dependent jobs) 
- This encourages a slightly different use case for Chronos which would benefit greatly from tasks not being disabled if they are not scheduled to run through the cron. 

I would love any feedback here! Does this seem like a use case we would want to support with Chronos? Thanks!

PR #374, opened 2015-02-08
Task flow parameters

jfenc91:
I have added two things here. 
  1) Task Flow Parameters 
  2) A triggered job type

1) I am trying to use chronos to kick off a long running chain of tasks that are performed on demand with set of parameters. There was a recent addition to add arguments to the trigger endpoint. However, these parameters were not propagated to dependent jobs (for good reason). Additionally, because of how there were implemented, the length of the arguments were limited to around 230 characters. 

To address this, I added a task flow state that gets persisted in zookeeper. It basically holds a list of environment variables to allow specialization of the arguments for each job on a per task basis. Additionally, I added a new endpoint to support receiving this parameter. Basic usage is as follows:
curl -L -X POST chronosNode:8080/scheduler/job/trigger/JobName -d '{"env":{"var1":"val1", "var2":"val2"}, "id":"UniqueTaskFlowId"}'

A job that has been configured with arguments such as: "--opt1 ${var1} --opt2 ${var2}" will run with  "--opt1 val1 --opt2 val2"

2) This feature starts to encourage a different usage pattern than is offered by a schedule based job. Basically, I didn't want my jobs getting disabled because they were not scheduled to run again. So, I added a job type that isn't scheduled and just waits to be triggered. This type is fully supported by the scheduler and api, but I have yet to start adding it to the web ui.

One thing that I did change is that the character ":" is no longer supported in the arguments set in the old trigger api. There are two solutions to this. First, we could flip the order of the order of the arguments and task flow id in the mesos task_id. Secondly, we could remove support for the arguments provided with the trigger api. That feature was added very recently and is possibly made redundant by these changes. 

 I would appreciate any feedback! Thanks! 
Also here is the feature request I made initially: https://github.com/mesos/chronos/issues/367

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
@@ -5,11 +5,11 @@ import javax.ws.rs._
 import javax.ws.rs.core.Response.Status
 import javax.ws.rs.core.{MediaType, Response}
 
+import com.codahale.metrics.annotation.Timed
+import com.google.inject.Inject
 import org.apache.mesos.chronos.scheduler.config.SchedulerConfiguration
 import org.apache.mesos.chronos.scheduler.graph.JobGraph
 import org.apache.mesos.chronos.scheduler.jobs._
-import com.codahale.metrics.annotation.Timed
-import com.google.inject.Inject
 import org.joda.time.{DateTime, DateTimeZone}
 
 import scala.collection.mutable.ListBuffer
@@ -84,6 +84,36 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
                   case _ =>
                 }
             }
+          case t: TriggeredJob =>
+            children.foreach {
+              child =>
+                jobGraph.lookupVertex(child).get match {
+                  case childJob: DependencyBasedJob =>
+                    val newChild = new TriggeredJob(
+                      name = childJob.name,
+                      command = childJob.command,
+                      epsilon = childJob.epsilon,
+                      successCount = childJob.successCount,
+                      errorCount = childJob.errorCount,
+                      executor = childJob.executor,
+                      executorFlags = childJob.executorFlags,
+                      retries = childJob.retries,
+                      owner = childJob.owner,
+                      lastError = childJob.lastError,
+                      lastSuccess = childJob.lastSuccess,
+                      async = childJob.async,
+                      cpus = childJob.cpus,
+                      disk = childJob.disk,
+                      mem = childJob.mem,
+                      disabled = childJob.disabled,
+                      softError = childJob.softError,
+                      uris = childJob.uris,
+                      highPriority = childJob.highPriority
+                    )
+                    jobScheduler.updateJob(childJob, newChild)
+                  case _ =>
+                }
+            }
         }
       }
       jobScheduler.sendNotification(job, "[Chronos] - Your job '%s' was deleted!".format(jobName))
@@ -119,6 +149,34 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
     }
   }
 
+
+  
+  @Path(PathConstants.triggerJobPatternPath)
+  @POST
+  @Timed
+  def triggerPost(@PathParam("jobName") jobName: String, 
+                  flowState: TaskFlowState
+               ): Response = {
+    try {
+      log.info(s"triggered with ${flowState.env mkString " "}")
+      require(flowState.id != null, s"Job $jobName must be given a unique task flow state id")
+      require(jobGraph.lookupVertex(jobName).isDefined, "Job '%s' not found".format(jobName))
+      val job = jobGraph.getJobForName(jobName).get
+      log.info("Manually triggering job:" + jobName)
+
+      jobScheduler.taskManager.enqueue(TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0, Some(flowState.id)), job.highPriority, flowState)
+      Response.noContent().build
+    } catch {
+      case ex: IllegalArgumentException =>
+        log.log(Level.INFO, "Bad Request", ex)
+        Response.status(Response.Status.BAD_REQUEST).entity(ex.getMessage)
+          .build()
+      case ex: Exception =>
+        log.log(Level.WARNING, "Exception while serving request", ex)
+        Response.serverError().build
+    }
+  }
+  
   @Path(PathConstants.jobPatternPath)
   @PUT
   @Timed
@@ -126,6 +184,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
               @QueryParam("arguments") arguments: String
                ): Response = {
     try {
+      //log.info("post args" + request.arguments mkString(" "))
       require(jobGraph.lookupVertex(jobName).isDefined, "Job '%s' not found".format(jobName))
       val job = jobGraph.getJobForName(jobName).get
       log.info("Manually triggering job:" + jobName)
@@ -214,5 +273,5 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
         throw new WebApplicationException(Status.INTERNAL_SERVER_ERROR)
     }
   }
-
 }
+
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/state/MesosStatePersistenceStore.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/state/MesosStatePersistenceStore.scala
@@ -2,10 +2,11 @@ package org.apache.mesos.chronos.scheduler.state
 
 import java.util.logging.{Level, Logger}
 
-import org.apache.mesos.chronos.scheduler.config.SchedulerConfiguration
-import org.apache.mesos.chronos.scheduler.jobs._
 import com.google.inject.Inject
 import org.apache.curator.framework.CuratorFramework
+import org.apache.mesos.chronos.scheduler.config.SchedulerConfiguration
+import org.apache.mesos.chronos.scheduler.jobs._
+import org.apache.mesos.chronos.utils.taskFlowSerde
 import org.apache.mesos.state.{InMemoryState, State}
 
 import scala.collection.mutable
@@ -28,6 +29,7 @@ class MesosStatePersistenceStore @Inject()(val zk: CuratorFramework,
   //TODO(FL): Add proper retry logic into the persistence.
   val jobPrefix = "J_"
   val taskPrefix = "T_"
+  val flowPrefix = "F_"
 
   // There are many jobs in the system at any given point in time.
   val jobName = (name: String) => "%s%s".format(jobPrefix, name)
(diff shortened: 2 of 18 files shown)


jfenc91:
@brndnmtthws What are your thoughts on this addition?

brndnmtthws:
This is a pretty big PR, so it will take some time to review.

CC @elingg

jfenc91:
Thanks for taking a look. I really appreciate it!

elingg:
Yes, I will also take a look when I have a chance. Thanks @brndnmtthws, @jfenc91

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -84,6 +84,36 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
                   case _ =>
                 }
             }
+          case t: TriggeredJob =>
```
elingg: The workaround here would be to be schedule a job far in the future. I'd rather not add an additional type here unless necessary. Would this workaround be acceptable?

src/main/scala/org/apache/mesos/chronos/scheduler/state/MesosStatePersistenceStore.scala
```diff
@@ -59,11 +64,26 @@ class MesosStatePersistenceStore @Inject()(val zk: CuratorFramework,
     }
   }
 
+  def removeTaskFlowState(id: String): Boolean = {
+    remove(flowName(id))
+  }
+  
+  def getTaskFlowState(id: String): TaskFlowState = {
```
elingg: I'm not fully convinced that adding task flow parameters to Chronos is the way to go here. Perhaps in your job configuration, you could create the environment variables for your job and make sure the proper environment variables in your chain of jobs are intact when you define your jobs. This could be done through scripting and you could utilize zookeeper or another key value store for this. The issue with adding this patch to Chronos is it adds another layer of dependency to zookeeper and also more complexity to setting the job arguments. One alternative thought is that we could add an inherit environment flag so child jobs can inherit their parents environment if the flag is set. Let's discuss!

jfenc91: Thanks @elingg for taking a look at my changes. 

You are correct, there are many ways to implement this. As you mentioned, an alternate way would be to change all the jobs to execute commands like: set_env && exec job where set_env would go out to a key value store. The question here becomes what is my primary key and what is responsible for updating the key value store. 

One option is to make the primary key the name of the job. However, this would make it impossible to ever run two instances of the same task flow simultaneously. Admittedly for cron jobs restarting the job while the previous one is still running is most likely a bad idea. However, for jobs that are triggered and parameterized it starts to make sense to allow multiple instances of a task flow to run simultaneously given there are resources available. (this is also a requirement for me)  

Alternatively a task flow id could be the primary key. This task id would have to be propagated and managed by chronos given that I do not want to replicate the scheduler logic elsewhere. This id would probably have to be an argument to the trigger command in order for the parameters to be properly set up before the job starts. While this works, I am not enthusiastic about it because it adds complexity to the requester and, by this point, it almost seems like I should be able to add parameters too.

I am all for allowing child jobs to inherit their parent environments. That is what I was trying to get at with this feature. However, I hit the 255 character limit on the mesos task id. Were you thinking of an alternate way of doing this? 

What is the hesitancy behind adding another layer of dependency on zookeeper? While zookeeper has its drawbacks, from my understanding, chronos will always need some sort of replicated store to work as a job scheduler. I don’t believe this addition will make it any harder to substitute in a different key/value store if that is desired. Additionally I don’t see parameters as being too big to store in zookeeper. What am I missing here? 

Also, for some perspective, it seems that another scheduler Azkaban supports "flow parameters”: http://azkaban.github.io/azkaban/docs/2.5. This doesn’t necessarily make it any any better or worse of an idea for chronos to support them too.

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
jfenc91: At some point two job types become semantically different enough that they should be separated. I agree, with this pull request on its own, I may as well just schedule a job for next the next century. However, if the ability to schedule multiple instances of the same job gets added (see the other comment) I think we should probably have another job type.

elingg:
I don't think providing a key value store for setting up env variables should be inside chronos, but should be done as a configuration before creating your chronos jobs.

Regarding creation of task flows and allowing multiple flows to run simultaneously, this is a reasonable functionality. Perhaps this should be a separate PR from the environment variables.

About having child jobs inherit their parent env, I was thinking of setting env of all child jobs if parent flag is set to true. If parent flag is set to false, then we don't propagate the environment settings. Of course, you could again also have an outside script that's setting the env params in which case we don't need this functionality inside chronos per se. Let me know if I'm missing something.

Regarding this dependency on zookeeper, I'm just not convinced a key value store for env variables should be part of chronos. This is custom functionality and it will add more complexity to chronos code as well as more complexity added to the zookeeper depdendency.

Glad to discuss further. I think it would really help to separate this into two PR's. One for setting the env and one for the task flow and new job type.

@brndnmtthws will also be reviewing today. Thanks again for your contribution!

jfenc91:
@elingg I am still unclear about what your are thinking in regards to child jobs inheriting their parent env. Do you see a way we can do this without zookeeper? Would these be just the environment variables defined in the job configuration or could they be defined by the trigger request? 

In order to implement this outside of Chronos and keep the simple rest interface to the scheduler, I will basically have to write a service to replicate the trigger command. This seems really suboptimal. Also, this will still depend on chronos tracking a task flow id for me.  

I will be happy to split this up into multiple PRs once we get a good idea of what we want here. Thanks for your time!

elingg:
It is possible to set the env variables outside of chronos as a separate script that creates the job definitions. Decoupling this from chronos seems ideal to me, because right now the code is quite clean, you just set the value as part of your job configuration. For the inheritance option, if we were to merge this into chronos, I pictured this as setting the environmentVariables field to set the environment variables and adding a new field to the job definition maybe called inherit-parent-env. When this flag is set, all the child jobs will then inherit those variables. I haven't thought through this idea in too much detail so feel free to let me know if I'm missing something.

Task Flow parameters seems to be a separate PR. We will want to find a good way to set their env variables, but at least for me, it will simplify the review process separating it out.

Let's continue this discussion after we get @brndnmtthws' input, and thanks for your patience.

src/main/scala/org/apache/mesos/chronos/scheduler/state/PersistenceStore.scala
```diff
@@ -8,12 +8,33 @@ import org.apache.mesos.chronos.scheduler.jobs._
...
+  def persistTaskFlowState(state: TaskFlowState): Boolean
+  
+  /**
    * Persists a job with the underlying persistence store
    * @param job
    * @return
    */
   def persistJob(job: BaseJob): Boolean
-
+
```
brndnmtthws: Whitespace here.

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -126,6 +184,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
               @QueryParam("arguments") arguments: String
                ): Response = {
     try {
+      //log.info("post args" + request.arguments mkString(" "))
```
brndnmtthws: Either drop this, or use `log.debug()`.

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -214,5 +273,5 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
         throw new WebApplicationException(Status.INTERNAL_SERVER_ERROR)
     }
   }
-
 }
+
```
brndnmtthws: Whitespace.

brndnmtthws:
I don't think we need a separate triggered job. The same thing can be accomplished with a scheduled job, by using `R0`.

Passing the env vars as flow state is a good idea. I'm a bit worried about creating a new znode for every run of every job. How about we use the root job name as the ID rather than a unique ID? Using the env vars from the previous job run should be sufficient. A dependant job can just retrieve the vars from all the parents and merge them (we can do this naively, for now).

We should also add the job schedule as part of these env vars to the job flow. Something like `CHRONOS_JOB_SCHEDULED_AT` with the scheduled date, or `CHRONOS_JOB_TRIGGERED_AT` if the job was triggered via the API.

Please split this up into 1 PR per feature, so we can go through one at a time.

jfenc91:
Thanks for taking a look @brndnmtthws. 

R0 jobs are disabled after their first run so that they can not be triggered again. Would you be okay a PR to change this? 

How would you feel about a PR that would create a configuration option to allow the same job to be executed multiple times concurrently? This is why I went for a unique id for flow state's znode. 

In regards to merging variables from all the parents, I am not sure I follow your train of thought here. Are you suggesting we add the flow state to the job definition znode? Or, are you recommending we have a znode with flow state for each job instead of sharing the flow state in a znode named by a flow id or root job name? 

Thanks!

elingg:
From my perspective, changing R0 jobs so they aren't disabled would be reasonable and a much cleaner way to implement a "triggered job".

I believe what @brndnmtthws is referring to is saving a job's env variables in zookeeper and pulling that information based on the job id so it can be easily be retrieved by child jobs. He can elaborate further if I'm missing something. We both feel that passing env to child jobs is a very useful feature. We just want the optimal design and not to store more than what we need to in zookeeper.

If you could split this into separate reviews, one for triggered job, one for flow state, and passing the env variables to child jobs, that would make it much easier for us to review. 

Thanks again for your help here!

jfenc91:
Got it.  Job id doesn't work if we start allowing multiple tasks from the same job definition to run simultaneously. We could do jobId-flowId, but I think that is what you all were trying to avoid.

elingg:
Indeed. With different PR's, we can isolate these different features and their interaction. 

About design, my vote is to review the feature for passing env variables to child jobs and the feature to not disable the R0 jobs in a clean way. TaskFlow I'm not yet convinced of. It sounds like a reasonable feature, but I want to make sure the design is clean and we're not writing to zookeeper too many times for example.

@brndnmtthws can comment if he has any other opinion.

jfenc91:
I have no interest in implementing a feature to allow for passing env variables to child jobs if there is no path in future PRs to make it work with multiple instances of a job running simultaneously. That said, I am all for more discussion and compromises! I could open empty PRs, or maybe this is a better use case for the issues section? Or, we could just continue here. If we decide this is a feature we don't want in main branch of chronos that is perfectly fine and I will move on. I really appreciate the time you all have spent with me discussing the addition of these features.

brndnmtthws:
A disabled job can always be triggered from the API.

elingg:
@brndnmtthws fair enough re R0 jobs. Doesn't seem like it needs a separate PR then.

jfenc91:
I could only get R0 jobs to go once. Am I missing something? https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala#L90-L91

brndnmtthws:
I wonder when that got changed. A disabled job should always be force-runnable.

brndnmtthws:
Actually, I just verified: you can (still) in fact trigger disabled jobs from the API. So there shouldn't be any issue there.

jfenc91:
I think that changed 11/9/14. With the commit message "Don't schedule queued jobs if they have been disabled". Although I should probably test again myself just to verify.

elingg:
@jfenc91, this has been idle for some time. Any objections to closing this? The open issue is still referencing this PR so our discussion will be saved here.

jfenc91:
Sure lets close this. I will break things up into smaller requests and kill TriggeredJob when I find the time. Thanks!

PR #375, opened 2015-02-12
Docs: remove note about jobs without schedules.

brndnmtthws:
cc @elingg


elingg:
Thanks for adding clarity to the docs! :+1:

PR #380, opened 2015-02-27
Enable checkpointing by default.

brndnmtthws:
cc @elingg @adam-mesos


elingg:
:+1: after the build passes

PR #383, opened 2015-03-04
expose more job variables via yaml files

mngo87:
Expose more variables to yaml file, to allow an user to have more control over job management.

### DIFF
--- a/bin/chronos-sync.rb
+++ b/bin/chronos-sync.rb
@@ -112,6 +112,7 @@ def normalize_job(job)
   newjob.delete 'lastSuccess'
   newjob.delete 'lastError'
   newjob.delete 'errorsSinceLastSuccess'
+  newjob.delete 'disabled'
 
   # Define optional fields, if not present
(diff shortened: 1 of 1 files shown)


elingg:
LGTM, any comments @brndnmtthws ?

brndnmtthws:
These fields were intentionally dropped. At the very least, it should ignore the `disabled` field because we don't want to re-enable jobs which were disabled due to failures (that defeats the purpose of the disable-after-N-failures feature).

The other fields were found to be more harm than good when modified, and thus were dropped. I'd be willing to include them back if necessary.

mngo87:
@brndnmtthws that makes sense in regards to disabled, I can remove that field.  The main fields that I want to update are epsilon and retries.  What are your thoughts on the other fields?  I don't see any harm keeping the async field as well.

brndnmtthws:
Yeah, it's reasonable. These things may be dropped in future releases (particularly the async mode), just FYI.

PR #384, opened 2015-03-06
[chronos-sync] Undefined fields should be ignored.

brndnmtthws:
If a job is missing fields which aren't mandatory, they should just be
ignored when comparing jobs.

cc @elingg


elingg:
Looks great, thx! :+1:

PR #387, opened 2015-03-11
Add state (i.e. queued, running) tracking for a job lifecycle. Currently...

mngo87:
..., everything is stored in memory. Look into persisting this state onto Cassandra.

PR #388, opened 2015-03-12
REST endpoint to retrieve job status

mikkokupsu:
The company that I work for uses Chronos to schedule pipelines and we needed an ability to get task status similarly as Mesos shows them; namely needed the ability see when job is running. I've created a new REST endpoint to retrieve this information as I didn't dare to update job related endpoints because I was quite sure if that could have broken something, at least is would have changed the entity content.

Feel free to ignore this pull request if you think your project does not need this.

Best regards,

Mikko Kupsu

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Also join us on IRC in #mesos on freenode.
 * [API](#api)
     - [Leader](#leaders)
     - [Listing Jobs](#listing-jobs)
+    - [Retrieving a Job status](#retrieving-a-job-status)
     - [Deleting a Job](#deleting-a-job)
     - [Deleting All Tasks for a Job](#deleting-all-tasks-for-a-job)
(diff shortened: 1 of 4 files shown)


brndnmtthws:
This seems to have some overlap with #387. We should try and reconcile the two PRs.

Also, it might make sense to pull the data out of Cassandra, rather than keeping it in memory. The downside is that it then requires C*.  Perhaps use C* for historical data, then store a limited number of records in memory?

elingg:
hello @mikkokupsu, any updates on this? This has been idle for some time so I wonder if it is safe to close. We could also reopen if there are further updates. Let me know. Thanks!

elingg:
hi @mikkokupsu, @brndnmtthws, Closing for now, feel free to reopen if you would like to continue to review process!

PR #391, opened 2015-03-14
Add webui_url to FrameworkInfo

guenter:
Tests pass, and tested setting the commandline option.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
     <joda-convert.version>1.7</joda-convert.version>
     <joda-time.version>2.3</joda-time.version>
     <lz4.version>1.2.0</lz4.version>
-    <mesos.version>0.20.1</mesos.version>
+    <mesos.version>0.21.0</mesos.version>
     <mesos-utils.version>${mesos.version}-1</mesos-utils.version>
     <metrics.version>3.1.0</metrics.version>
     <raven.version>4.1.2</raven.version>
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
```diff
@@ -48,17 +50,29 @@ class MainModule(val config: SchedulerConfiguration) extends AbstractModule {
...
-    val frameworkInfo = FrameworkInfo.newBuilder()
+    val frameworkInfoBuilder = FrameworkInfo.newBuilder()
       .setName(config.mesosFrameworkName())
       .setCheckpoint(config.mesosCheckpoint())
       .setRole(config.mesosRole())
       .setFailoverTimeout(config.failoverTimeoutSeconds())
       .setUser(config.user())
-    frameworkIdUtil.setIdIfExists(frameworkInfo)
-    frameworkInfo.build()
+
```
elingg: This is a nit, but let's but the else if and else should be one line up. See another example in the Chronos code:

//If the job was deleted after the taskId was added to the queue, the task could be empty.
      if (jobOption.isEmpty) {
        //remove invalid task
        removeTask(taskId)
        None
      } else if (jobOption.get.disabled) {
        None
      } else {
        Some(taskId, job)
      }

pom.xml
```diff
@@ -31,7 +31,7 @@
     <joda-convert.version>1.7</joda-convert.version>
     <joda-time.version>2.3</joda-time.version>
     <lz4.version>1.2.0</lz4.version>
-    <mesos.version>0.20.1</mesos.version>
+    <mesos.version>0.21.0</mesos.version>
```
elingg: Nice

elingg:
Just one style comment, but it looks great. :+1: Thanks for adding the web UI! Any comments @brndnmtthws?

brndnmtthws:
:+1:, merge away.

guenter:
@brndnmtthws @elingg thanks! Made the style change.

PR #396, opened 2015-03-19
Set the offer frequency to 5 secs

elingg:
When an offer is refused, we wait 5 secs before receiving that offer
again (consistent with Marathon). This will have an effect on triggered
jobs, in that they will be a short delay in triggering them. However,
it will ensure that Chronos doesn’t receive and decline offers too
frequently and that Chronos doesn’t receive more than it’s fair share
of offers.


elingg:
@brndnmtthws, please take a look at your earliest convenience, thx!

brndnmtthws:
:+1: 

Might be worth making it a config param.

elingg:
@brndnmtthws, true, a config param may be a useful subsequent PR for those who need high latency in their triggered jobs or would like a more immediate response once a previously unavailable resource becomes available.

PR #397, opened 2015-03-19
Support disabling prompt for deleting missing jobs

sdwr98:
We have an automated script that generates chronos jobs and syncs them with Chronos.  We'd like to be able to delete missing jobs without the prompt.

### DIFF
--- a/bin/chronos-sync.rb
+++ b/bin/chronos-sync.rb
@@ -43,7 +43,7 @@ opts = OptionParser.new do |o|
     options.http_auth_user = cred_split[0].strip
     options.http_auth_pass = cred_split[1].strip
   end
-  o.on("--delete-missing", "Prompt to delete missing jobs") do
+  o.on("--delete-missing", "Delete missing jobs from chronos. Prompts for confirmation unless --force is also passed.") do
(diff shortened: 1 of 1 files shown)


elingg:
This seems like a reasonable option to me and looks good, although it is adding more complexity to the script. Ideally, we would modularize this script and break it down some. Any objections @brndnmtthws?

brndnmtthws:
How about just reuse the `-f` flag for this purpose?

sdwr98:
@brndnmtthws I actually thought the first time that I used it that `-f` implied `--delete-missing`.  What are your thoughts on making that change? Or I can make it so that -f replaces the no-prompt option.

brndnmtthws:
I think just updating the docs (and code) to reflect that `-f` means no prompt on delete should be sufficient.

sdwr98:
@brndnmtthws @elingg Updated to use -f instead of another flag. Thanks!

brndnmtthws:
Looks nice, thanks! Has it been tested?

sdwr98:
Yes, it has.  Without --force it retains the current behavior, with --force it just issues the DELETE.

PR #399, opened 2015-03-19
Update Readme so links are correct

elingg:


elingg:
Update links in readme, cc @brndnmtthws

brndnmtthws:
Looks good!

:+1:

PR #400, opened 2015-03-24
Upgrading Chaos to 0.6.3 (env variables for http auth and ssl)

swartzrock:


pom.xml
```diff
@@ -65,7 +65,8 @@
     </repository>
     <repository>
       <id>libs-folder</id>
-      <url>file://${basedir}/lib</url>
+      <name>libs-folder</name>
+      <url>file:${project.basedir}/lib2</url>
```
elingg: The lib folder should just be named lib, no need to rename it.

elingg:
LGTM, just one comment to be addressed.

swartzrock:
Thanks for catching that error! Now fixed to just "lib". Awesome!

elingg:
:+1:

PR #402, opened 2015-03-24
P/mike/job history upstream

mngo87:


mngo87:
successful_schedule_job
running_job
no_past_data_job
ui_overflow
expanded_job_view
cassandra_data_outage

mngo87:
The purpose of this feature is to show the past task executions for each job (if data exists).  Data is extracted from the existing Cassandra setup.  The tasks are reverse sorted by date, with the most recent at the top.

There are two commits for this request.  The commits are separated to show the initial change to the cassandra table schema by changing the indexing scheme.  To adapt to the existing schema that upstream uses, I changed to go from a partition key to a secondary index.  After getting feedback, I'll squash the commits and merge the final version.

mngo87:
@brndnmtthws @elingg - since this is the same area as the last pull request, can you please review again? thanks :)

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobStatWrapperSerializer.scala
```diff
@@ -0,0 +1,80 @@
...
+      json.writeFieldName("status")
+      json.writeString(taskStat.taskStatus.toString())
+
+      json.writeEndObject()
+    }
+    json.writeEndArray()
+
+    json.writeEndObject()
+  }
+}
```
brndnmtthws: No newline at EOF.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -80,6 +80,134 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+   */
+  private def getJobStatsByName(jobName: String, numTasks: Int): Option[List[Row]] = {
+    var rowsListFinal: Option[List[Row]] = None
+    try {
+      getSession match {
+        case Some(session: Session) =>
+          val query = s"SELECT * FROM ${config.cassandraTable()} WHERE ${JOB_NAME}='${jobName}';"
+          val prepared = statements.getOrElseUpdate(query, {
+            session.prepare(
+              new SimpleStatement(query)
```
brndnmtthws: Maybe set the consistency level here? Like this: https://github.com/mesos/chronos/blob/4a7319edcda1a087b942d35c2efd753c7cea0eed/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala#L296-L297

brndnmtthws:
Looks great, minus some minor comments. Can you also squash & rebase into 1 commit, please.

Thanks!

brndnmtthws:
Another thing: can you update the README to explain how this works? I'm worried some folks (who don't use the C* stuff) will be confused as to why they don't have this data.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
mngo87: Looks like consistency level settings only apply to writes.  I get this exception: com.datastax.driver.core.exceptions.InvalidQueryException: ANY ConsistencyLevel is only supported for writes

mngo87: This doesn't make sense.  Let me try something else.

brndnmtthws: Hmm yeah there should be a consistency setting for both reads and writes.

mngo87: Still getting the problem.  What version of cassandra are you using? I am using 2.0.12.

Looked at the datastax repo: https://github.com/datastax/java-driver.  Seems like it is happening somewhere.  

The apache cassandra repo also has a code stop for it here: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/ConsistencyLevel.java#L306

brndnmtthws: Oh, I see now.

This is right: you can't use `ANY` with reads, only writes.

What you could do is have a check to see if it's set to `ANY`, and then use `ONE` instead, otherwise go with the specified value. Does that make sense?

mngo87: Sounds good.  Though according to this link (https://www.datastax.com/documentation/cassandra/2.0/cassandra/dml/dml_config_consistency_c.html), I don't see any reference as to why ANY is not allowed for reads.

brndnmtthws: C* treats read/write consistency differently. Check the table at https://www.datastax.com/documentation/cassandra/2.0/cassandra/dml/dml_config_consistency_c.html for details. Note that there's no `ANY` for reads.

The `ANY` feature is implemented with hinted handoff, which doesn't make much sense for reads.

mngo87: doh. i read "ALL" and "ANY" as the same thing.  thanks for pointing that out.

mngo87:
@brndnmtthws - comments addressed. can you review again?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -80,6 +80,143 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+   */
+  private def getJobStatsByName(jobName: String, numTasks: Int): Option[List[Row]] = {
+    var rowsListFinal: Option[List[Row]] = None
+    try {
+      getSession match {
+        case Some(session: Session) =>
+          val query = s"SELECT * FROM ${config.cassandraTable()} WHERE ${JOB_NAME}='${jobName}';"
+          val prepared = statements.getOrElseUpdate(query, {
+            var level = ConsistencyLevel.ANY
+            if (ConsistencyLevel.ANY.name().equalsIgnoreCase(config.cassandraConsistency())) {
```
brndnmtthws: Maybe pull this check out into a separate method like `readConsistencyLevel()`? Then have:

```scala
            session.prepare(
              new SimpleStatement(query)
                .setConsistencyLevel(readConsistencyLevel())
                .asInstanceOf[RegularStatement]
            )
```

brndnmtthws:
One minor nit pick.

Any comments from @elingg?

elingg:
Thanks for the contribution @mngo87 and for the review @brndnmtthws, I'll take a final look at this ASAP.

mngo87:
Sure no problem.  I updated once more to address @brndnmtthws 's comments

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -310,4 +447,15 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
     }
 
   }
+
+  private def readConsistencyLevel() : ConsistencyLevel = {
+    var level = ConsistencyLevel.ANY
```
brndnmtthws: You can omit this `level` var altogether. The last statement in the function will be the return value.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -310,4 +447,15 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+    if (ConsistencyLevel.ANY.name().equalsIgnoreCase(config.cassandraConsistency())) {
+      //reads do not support ANY
+      level = ConsistencyLevel.ONE
```
brndnmtthws: This line should just be:

```scala
      ConsistencyLevel.ONE
```

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -310,4 +447,15 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+    } else {
+      level = ConsistencyLevel.valueOf(config.cassandraConsistency())
```
brndnmtthws: Similar to above.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -310,4 +447,15 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+    }
+    level
```
brndnmtthws: Drop this line.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
mngo87: good point. will update

brndnmtthws:
Looks great now, thanks.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -80,6 +80,136 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+              cDefs.contains(SLAVE_ID) &&
+              cDefs.contains(IS_FAILURE)) {
+            var jobName = row.getString(JOB_NAME)
+            var taskId = row.getString(TASK_ID)
+            var taskTimestamp = row.getDate(TIMESTAMP)
+            var taskState = row.getString(TASK_STATE)
+            var slaveId = row.getString(SLAVE_ID)
+            var isFailure = row.getBool(IS_FAILURE)
+
+            var taskStats = taskMap.getOrElseUpdate(taskId, new TaskStat(taskId, jobName, slaveId))
```
elingg: nit: this variable should probably be named taskStat

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -80,6 +80,136 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+
+  /**
+   * Returns most recent tasks by jobName and returns only numTasks
+   * @param jobName
+   * @param numTasks
+   * @return returns a list of past and currently running tasks,
+   *         the first element is the most recent.
+   */
+  def getParsedJobStatsByName(jobName: String, numTasks: Int): List[TaskStat] = {
+    val taskMap = Map[String, TaskStat]()
```
elingg: Why can't we persist the information for this Map in Cassandra and load it? If this method is called at different times we may get a TaskStat with a start date but no end date or a TaskState with an end date but no start date.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
```diff
@@ -80,6 +80,136 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
+      }
+    }
+    !compareAscDate
+  }
+
+  /**
+   * Returns most recent tasks by jobName and returns only numTasks
+   * @param jobName
+   * @param numTasks
+   * @return returns a list of past and currently running tasks,
```
elingg: If we could clean up this method simply update the rows that need to be modified, that will be much cleaner. Different fields of TaskStat are being persisted and loaded at different times and the way the object is being manipulated is not clear.

elingg:
This is really great work, @mngo87! My main concern at this point is with clean code, specifically in the method, getParsedJobStatsByName. I want to make sure we have a clean way of updating the appropriate fields in TaskStat and loading and saving the object to DB. Right now we modify different fields at different times, which is not ideal and difficult to read.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
mngo87: The reason I did not go down the route of persisting this map is that now there will be two sources of information to update.  Every time an insertion is done to the existing cassandra table (i.e. chronos table), another update will have to be done to this aggregate taskMap data in cassandra.  I was worried more bugs may result from that.  Hence, I opted to have one source of truth, which is the existing data storage setup.

In regards to having incomplete data (no start date and/or no end date), I can only see a situation in which there is a start date but no end date (i.e. task still running).  If the reverse happens, I would argue that the bug happens further up the chain.

Thoughts?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
mngo87: To be clear there is no row update being done here.  This is merely a read from the cassandra table and parsing the data into the TaskStat object.  Once the UI retrieves this data, it is thrown away.

mngo87:
Thanks @elingg for the review.  Let me know your thoughts on what I said.  Just to re-iterate, TaskStat is just merely a object representation of the chronos table.  It will pull all the data at once and then parse.  It will not get the subsequent updates during parsing.  That can be a feature for later.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
mngo87: hah good catch.  will update.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
elingg: Hm, just to clarify, the issue I see here is that let's say you first call this method when the task starts running. You create a TaskStat object with start date. Then later you call this method when the task finishes, you have created a TaskStat object with end date. I guess TaskStat is a throw away object that simply updates the UI, but I feel that we should have one object that contains the full start and end date and also we should possibly store is somewhere. If it is a throw away object, perhaps instead we do away with it and simply call updateStartDate and updateEndDate which updates the UI. I feel this method is a bit confusing. It seems you are correct that there is no bug, but it is hard to read. Better readability and simplicity would be appreciated.

elingg:
Thanks for the quick response @mngo87! I left by reply above. That's fine that we don't persist the TaskStat or the Map, but I would like to clean up that method for simplicity and readability.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobStats.scala
mngo87: Maybe I'm not following you completely.  

Yes it is inefficient in the sense that the TaskStat object is created/deleted every time this method is invoked.  So the TaskStat object represents a whole task, hence it would contain both the full start and end date.  In terms of storing taskMap in memory and merely updating the timestamps (if they change), I like that idea.  However, it doesn't give a large benefit in performance, cause the chronos table should be queried regardless to get the correct data.  Hence, the parsing logic (which is dirty) will still have to be there. Do you have any ideas on how to make this easier to read?

elingg: Hm, without confusing ourselves further, I think storing the TaskStat Map in memory and removing TaskStat's from the Map once tasks are completed will simplify the code and make it more understandable. Would that option work for you? The parsing logic will still be there, but maybe we can break it down into two methods to make it cleaner?  One method could be reading the rows and the other method could be updating the map.

mngo87: Sure I can separate out.  Though if I start removing items from the map that may complicate the logic.  For example the reconciliation feature makes it hard to determine when a task is actually finished.

elingg: I think we could clear the map at a certain point in time to account for that. Would that work? The other option is storing the map in Cassandra instead of in memory. Yes, I think breaking down the method is a great idea.

mngo87:
@elingg - I just updated the request.  Instead of dealing with when to clear the taskMap, I decided to go back to the disposable map model and use it for temporary storage.  i separated out the code.  Let me know what you think

elingg:
@mngo87, with the cleaner logic, the disposable map makes more sense. I think it is a reasonable design. Nice work! I just made one comment to address, but it LGTM

mngo87:
@elingg - let me know your thoughts on my comments and I'll address if needed.  Thanks for the review.

elingg:
Responded, thanks!

mngo87:
@elingg - updated and tested.

elingg:
Sweet! :100:

PR #404, opened 2015-03-25
Updating Slack integration to comply with new Slack webhook format

sdwr98:
Slack has changed their web hook format, this update makes the Slack integration work again.


brndnmtthws:
Looks fine to me.

What happened to the extra parameters (like channel)?

Any comments @elingg?

sdwr98:
Slack changed their format - each web hook URL is specific to a channel and account, so the extra parameters (channel, token) are now merged into the URL.

brndnmtthws:
Understood, thanks for clarifying.

elingg:
:+1:  I had the same question, but this makes sense. Let's merge!

PR #405, opened 2015-03-27
[BUILD] - Updated to mesos-0.22.0

clehene:
separated version for mesos-utils (to 0.21.0-1) as they are in different repos

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>org.apache.mesos</groupId>
   <artifactId>chronos</artifactId>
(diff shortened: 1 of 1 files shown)


ConnorDoyle:
@clehene let's fix the POM first.  We don't need to declare a dependency on the mesos artifact since it is pulled in transitively with mesos-utils. (see https://github.com/mesos/chronos/issues/406)

clehene:
@ConnorDoyle K. closing

ConnorDoyle:
@clehene, a 0.22.0 version of `mesos-utils` is now available: see https://github.com/mesosphere/mesos-utils/releases/tag/v0.22.0-1

PR #408, opened 2015-03-27
[DOC] - Fixed launch example in README

clehene:


elingg:
I'm not convinced we need to change this instruction. java -cp target/chronos*.jar org.apache.mesos.chronos.scheduler.Main --master zk://localhost:2181/mesos --zk_hosts localhost:218 works successfully for me.

clehene:
@elingg it may have been just my setup. Closing

PR #410, opened 2015-03-30
Adding subject to Slack notification

sdwr98:
Whoops - should have added this in the last PR.  Having the subject in the slack notification helps distinguish what environment the alert is coming from.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/notification/SlackClient.scala
+++ b/src/main/scala/org/apache/mesos/chronos/notification/SlackClient.scala
@@ -19,9 +19,13 @@ class SlackClient(val webhookUrl: String) extends NotificationClient {
 
     // Create the payload
     generator.writeStartObject()
-
+    
     if (message.nonEmpty && message.get.nonEmpty) {
-      generator.writeStringField("text", message.get)
+      if (subject != null && subject.nonEmpty) {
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/mesos/chronos/notification/SlackClient.scala
```diff
@@ -21,7 +21,7 @@ class SlackClient(val webhookUrl: String) extends NotificationClient {
     generator.writeStartObject()
 
     if (message.nonEmpty && message.get.nonEmpty) {
```
elingg: Just a quick question. Did you want to check for subject.nonEmpty?

sdwr98: I was following MailClient and RavenClient, both of which seem to assume that the subject is present.  I can add another check, however, if you would like.

elingg: I think an additional check for an error condition could not hurt. If you could add that, it would be great.

sdwr98: Done - thanks for the feedback!

elingg:
LGTM

PR #414, opened 2015-04-01
Remove chronos-sync skip logic

sdwr98:
The chronos-sync.rb script currently has logic in it to skip jobs ahead if they have a schedule in the past.  However, this logic is broken for irregular intervals (P1M, for example).  As you can see below, it incorrectly sets the new date to 2015-04-02 instead of 2015-04-01. 

    Schedule for 'build_gas_stubs' begins in the past!  Resetting to from R/2014-09-01T11:05:00-04:00/P1M to R/2015-04-02T11:05:00.000-04:00/P1M

Chronos itself already has logic to do this (in JobUtils.makeScheduleStream), so I'm not sure if it's still necessary to do so in the sync script.

### DIFF
--- a/bin/chronos-sync.rb
+++ b/bin/chronos-sync.rb
@@ -302,16 +302,6 @@ if !options.skip_sync
           end
         end
         if options.force || !scheduled_jobs.include?(name) || normalize_job(existing_job).to_a.sort_by{|x|x[0]} != normalize_job(new_job).to_a.sort_by{|x|x[0]}
-          # Check if scheduled start time is in the past
-          start_time = DateTime.iso8601(/^R\d*\/([^\/]+)\//.match(new_schedule)[1])
-          if start_time < cur_datetime
-            split_schedule = new_schedule.split(/\//)
-            # Reset start time to the next available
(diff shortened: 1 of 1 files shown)


elingg:
It would be nice to fix this script to work correctly, but this is a reasonable fix in the meantime. Maybe you could add a TODO to skip jobs in the past and mention that the logic was removed since it had bugs. Any comments @brndnmtthws ?

brndnmtthws:
It seems okay to me. I don't remember exactly why I added this, but it seems like there would have been a reason. Have you tested that it exhibits the expected behaviour when you remove the code?

sdwr98:
It does - there's a message from JobUtils when you submit the job to that effect:

    Skipped forward 8 iterations, modified start from 'Monday, September 1, 2014' to 'Friday, May 1, 2015

The only thing that I noticed is that before the first job runs, the schedule on the job is still the original date from the sync, not the skipped-forward version.  But after the first run it gets update. 

If you'd prefer me to take a pass at fixing the sync script I will, I'll have to dust off the Ruby skills :)

brndnmtthws:
My only concern is all of the cycles wasted in `calculateSkips`, but I guess it's not a big deal.

:+1:

PR #425, opened 2015-04-09
Add API endpoint to update elements processed for a particular task and ...

mngo87:
...UI to read that value

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
@@ -71,6 +71,8 @@ trait BaseJob {
   def arguments: Seq[String] = List()
 
   def softError: Boolean = false
+
+  def dataProcessingJobType: Boolean = false
 }
 
 @JsonDeserialize(using = classOf[JobDeserializer])
@@ -101,7 +103,8 @@ case class ScheduleBasedJob(
                              @JsonProperty override val environmentVariables: Seq[EnvironmentVariable] = List(),
                              @JsonProperty override val shell: Boolean = true,
                              @JsonProperty override val arguments: Seq[String] = List(),
-                             @JsonProperty override val softError: Boolean = false)
+                             @JsonProperty override val softError: Boolean = false,
+                             override val dataProcessingJobType: Boolean = false)
   extends BaseJob
 
 
@@ -132,5 +135,6 @@ case class DependencyBasedJob(
                                @JsonProperty override val environmentVariables: Seq[EnvironmentVariable] = List(),
                                @JsonProperty override val shell: Boolean = true,
                                @JsonProperty override val arguments: Seq[String] = List(),
-                               @JsonProperty override val softError: Boolean = false)
+                               @JsonProperty override val softError: Boolean = false,
+                               override val dataProcessingJobType: Boolean = false)
   extends BaseJob
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Also join us on IRC in #mesos on freenode.
     - [Adding a Scheduled Job](#adding-a-scheduled-job)
     - [Adding a Dependent Job](#adding-a-dependent-job)
     - [Adding a Docker Job] (#adding-a-docker-job)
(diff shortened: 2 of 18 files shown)


mngo87:
Here is the UI.  There are a couple options on display.  If a job has been marked to track data (elements processed) then the UI will show a count if it exists.  If a job has not been marked to track data, then no UI is displayed for that field.  See below.

Job does not track data.  Same display.
nodataflag_ui

Job does track data, but no count values to display.
dataflagon_nodata_ui

Job does track data and there is some count values to display.
dataflagon_withdata_ui

mngo87:
@elingg - please review

mngo87:
@brndnmtthws - can you also take a look?

brndnmtthws:
Looks good to me.

elingg:
@mngo87, let me get back to you ASAP. Today was a busy day.

elingg:
My first comment is shouldn't it be 2/total items to process or 4/total items to process. 2 or 4 is confusing without a total number of items to process.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
```diff
@@ -101,7 +103,8 @@ case class ScheduleBasedJob(
                              @JsonProperty override val environmentVariables: Seq[EnvironmentVariable] = List(),
                              @JsonProperty override val shell: Boolean = true,
                              @JsonProperty override val arguments: Seq[String] = List(),
-                             @JsonProperty override val softError: Boolean = false)
+                             @JsonProperty override val softError: Boolean = false,
+                             @JsonProperty override val dataJob: Boolean = false)
   extends BaseJob
```
elingg: Thanks for adding this at the job level!

elingg: Should we call this dataProcessingType instead of dataJob. I think that would be more descriptive

elingg:
@mngo87, Nice work. I made a few comments. Could you address them before I take a second look? Thanks!

mngo87:
@elingg in regards to the total items processed.  Sometimes a job may or may not know the total items.  Hence, I did not think it was necessary.  I can add an optional field to track that if desired, but in Flixster's use case we did not find this information as helpful.  Instead we relied on historical counts to track correctness.  For example, if jobA processed ~10,000 items every time and one time processed 5,000 items.  Then we would investigate why there was an execution with 5,000 items.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
mngo87: I could add it as a type, but this field would now be an enum or a string.  Otherwise, I would not know what true or false meant.  @elingg thoughts?

elingg: My main concern here is in regards to naming. I think it should still be a field of Job, but the name "dataProcessingType" or "dataProcessingJobType" is more descriptive to me than "dataJob"

elingg:
@mngo87, I see, that's fair to leave total items off for now. Could the label be "Items Processed" instead of "Processed"? That is more descriptive.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
mngo87: yup agreed that the naming could be better.  will update.

elingg: cool thanks

mngo87:
@elingg - naming question for you.  My API calls everything elements.  So should I label as "Elements Processed" instead?

elingg:
Sure! Elements Processed also sounds like a good general description for Data Processing.

mngo87:
@elingg - please take a look again

elingg:
Looks great. Can you confirm these final changes were tested? If so, let's merge

mngo87:
Yup tested it before I messaged you.

elingg:
Yup, makes sense.

PR #427, opened 2015-04-10
Adding description and owner name fields to jobs

aimeemorgan:
This commit adds backend support for job description and owner name fields, plus support for viewing and editing these fields via the Chronos web UI.

Here's a screenshot showing the UI changes:
chronos_ui

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,6 +374,7 @@ you can also use a url in the command field, if your mesos was compiled with cUR
 | Field               | Description                                                                                              | Default                        |
 | ------------------- |----------------------------------------------------------------------------------------------------------| -------------------------------|
 | name                | Name of job.                                                                                             | -                              |
+| description         | Description of job.                                                                                      | -                              |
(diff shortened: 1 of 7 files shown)


mngo87:
@elingg - can you take a look at this too?

elingg:
LGTM! Any objections @brndnmtthws?

mngo87:
@elingg @brndnmtthws - just bumping so you don't forget.  thanks for reviewing.

elingg:
Looks like no objections from @brndnmtthws so we should be good. One last thing that I forgot to mention, @aimeemorgan. Could you please update the readme?

aimeemorgan:
@elingg Sure. Should I add that commit to this pull request, or open a new one?

elingg:
I would prefer if we also include that in this PR @aimeemorgan.

aimeemorgan:
@elingg Added a commit with the readme updates.

elingg:
:+1:

Issue #256, opened 2014-09-12
Support for constraints in chronos


mattomatic:
Is it possible to add support to chronos for constraints, such as those that marathon already supports (https://mesosphere.github.io/marathon/docs/constraints.html)?

For example, it would be very useful to be able to schedule jobs to be run in a particular data center. 

Also, is there an existing workaround for this?

elingg:
Adding constraints would be an important feature for Chronos. Patches from the community would be excellent if we aren't able to add this in our immediate roadmap.

veverjak:
Having marathon-like constrains feature would be super cool - https://mesosphere.github.io/marathon/docs/constraints.html

kutchar:
+1

mjbroekman:
+1 ... having the ability to specify parts of your cluster to run jobs on would great...

synergyhue:
+1

scaldmaster:
+1

siddharth178:
+1

freshmatrix:
+1

SEJeff:
Hey gents, instead of a pointless +1, just click the Subscribe button on the top right. More and more +1s are not very useful and just spam the developers unnecessarily.

ssk2:
@SEJeff, the +1s can be helpful for prioritisation. Ideally we'd have a less spammy system or the ability to vote built into GH issues, of course.

SEJeff:
That is what subscribe is for :)

I comaintain a few largish open source projects (> 500 unique contributors each), salt and graphite. It works pretty well.

gyllen:
@brndnmtthws How hard would this be to accomplish, do you have any idea. It is most certainly an important feature.

kapil-malik:
This works pretty well now for us in latest version. Curious why this issue is still open now?

potto007:
Isn't it still restricted to just the EQUALS constraint at this point? I
think the intent is to bring it to parity with Marathon wrt constraints.

On Thu, Jun 18, 2015, 7:53 PM KAPIL MALIK <notifications@github.com> wrote:

> This works pretty well now for us in latest version. Curious why this
> issue is still open now?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mesos/chronos/issues/256#issuecomment-113338593>.
>
-- 

Paul Otto
Principal Software Architect
Otto Ops LLC | *OttoOps.com*
970.343.4561 office
720.381.2383 cell

kapil-malik:
[...]

PR #431, opened 2015-04-20
Add constraints

wndhydrnt:
This PR adds support for `constraints` as described in #256.

The format to define constraints of a job is the same as in Marathon.
Note that currently only the `CLUSTER` constraint has been implemented.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
@@ -1,5 +1,6 @@
 package org.apache.mesos.chronos.scheduler.jobs
 
+import org.apache.mesos.chronos.scheduler.jobs.constraints.Constraint
 import org.apache.mesos.chronos.utils.JobDeserializer
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
@@ -77,6 +78,8 @@ trait BaseJob {
   def softError: Boolean = false
 
   def dataProcessingJobType: Boolean = false
+
+  def constraints: Seq[Constraint] = List()
 }
 
 @JsonDeserialize(using = classOf[JobDeserializer])
@@ -110,7 +113,8 @@ case class ScheduleBasedJob(
                              @JsonProperty override val shell: Boolean = true,
                              @JsonProperty override val arguments: Seq[String] = List(),
                              @JsonProperty override val softError: Boolean = false,
-                             override val dataProcessingJobType: Boolean = false)
+                             @JsonProperty override val dataProcessingJobType: Boolean = false,
+                             @JsonProperty override val constraints: Seq[Constraint] = List())
   extends BaseJob
 
 
@@ -144,5 +148,6 @@ case class DependencyBasedJob(
                                @JsonProperty override val shell: Boolean = true,
                                @JsonProperty override val arguments: Seq[String] = List(),
                                @JsonProperty override val softError: Boolean = false,
-                               override val dataProcessingJobType: Boolean = false)
+                               @JsonProperty override val dataProcessingJobType: Boolean = false,
+                               @JsonProperty override val constraints: Seq[Constraint] = List())
   extends BaseJob
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/EqualsConstraint.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/EqualsConstraint.scala
@@ -1,0 +1,19 @@
+package org.apache.mesos.chronos.scheduler.jobs.constraints
+
+import org.apache.mesos.Protos
+
+case class EqualsConstraint(attribute: String, value: String) extends Constraint {
+
+  def matches(attributes: Seq[Protos.Attribute]): Boolean = {
+    attributes.foreach { a =>
+      if (a.getName == attribute && a.getText.getValue == value) {
+        return true
+      }
+    }
+    false
+  }
+}
+
+object EqualsConstraint {
+  val OPERATOR = "EQUALS"
+}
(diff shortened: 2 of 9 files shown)


elingg:
Nice @wndhydrnt, will take a look ASAP!

elingg:
Looking good. One overall comment to begin wtih - could we name this constraints or job constraints? Cluster constraints make it seem like it's for all jobs in the cluster.

wndhydrnt:
The implementation is similar to how [Marathon defines constraints](https://mesosphere.github.io/marathon/docs/constraints.html).
The `ClusterConstraint` that has the operator `CLUSTER` is just one implementation of the basic trait `Constraint`. Looking at the constraints supported by Marathon, the `LIKE` constraint could be useful and can be added later.
From that point of view, I would not change the naming but the docs may need to provide more details.

However, it all depends on whether there will be other constraints that need to be implemented. If you think that this is not very likely, I would suggest the following:

- Remove the `CLUSTER` operator as it is kind of pointless in this case
- Use a hash as the value of `constraints` in the configuration of a job, like
```json
...
"constraints": {"rack": "rack-1", "profile": "high-load"},
...
```
- Implement a `Constraint` class instead of a trait

What do you think @elingg ?

elingg:
Ah, in Marathon cluster constraints are for cluster wide constraints, which I think we should also have for Chronos eventually: 

```CLUSTER allows you to run all of your app's tasks on slaves that share a certain attribute. This is useful for example if you have apps with special hardware needs, or if you want to run them on the same rack for low latency.```

It seems the only implementation you have here is cluster wide constraints, but these are implemented at the job level. I see this in your code:

```
  offerResources.toIterator.find { ors =>
+                  ors._2.canSatisfy(neededResources) && checkConstraints(ors._1.getAttributesList.asScala, job.constraints)
```




Cluster wide constraints should be not be set and checked at the job level.

**Comment by wndhydrnt** [comment]
I think I got your point. The difference is that a Marathon app can have several instances (= tasks) running at the same time where with Chronos there is only one task of a job running at the same time.

Here is an example, just to make clear what I want to achieve.
Given five Mesos slaves with the following attributes:

Slave | Attributes
------- | -------------
s0 | rack:a
s1 | rack:a
s2 | rack:b
s3 | rack:b
s4 | rack:c

And a job with `[["rack", "CLUSTER", "b"]]` set as the value of the `constraints` key.
Chronos should make sure that tasks of this job always run on `s2` or `s3` and never on `s0`, `s1` or `s4`.

---

I also thought of a different name for the operator instead of `CLUSTER`. Maybe something like `EQUAL` or `EQUALS` better reflects what the constraint is doing.

**Comment by elingg** [comment]
Yes, @wndhydrnt, this makes sense and I agree. I think the EQUALS operator would be appropriate here.

**Comment by wndhydrnt** [comment]
:+1: I'll change the naming and update the PR

**Comment by elingg** [comment]
Great, nice work! I love that we are putting the first round of constraints into Chronos.

**Comment by wndhydrnt** [comment]
Done. Please take a look at it again @elingg.

File: src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala [hunk 1]
```diff
@@ -110,7 +113,8 @@ case class ScheduleBasedJob(
                              @JsonProperty override val shell: Boolean = true,
                              @JsonProperty override val arguments: Seq[String] = List(),
                              @JsonProperty override val softError: Boolean = false,
-                             override val dataProcessingJobType: Boolean = false)
```
elingg: not your code, but could we add @JsonProperty to dataProcessingJobType while you are at it. Not sure why this was left off.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/EqualsConstraint.scala
```diff
@@ -0,0 +1,20 @@
...
+
+case class EqualsConstraint(attribute: String, value: String) extends Constraint {
+
+  def matches(attributes: Seq[Protos.Attribute]): Boolean = {
+    attributes.foreach { a =>
+      if (a.getName == attribute && a.getText.getValue == value) {
+        return true
+      }
+    }
+
```
elingg: Why do we have a space here?

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala
```diff
@@ -99,6 +100,16 @@ import scala.concurrent.ExecutionContext.Implicits.global
   def generateLaunchableTasks(offerResources: mutable.HashMap[Offer, Resources]): mutable.Buffer[(String, BaseJob, Offer)] = {
     val tasks = mutable.Buffer[(String, BaseJob, Offer)]()
 
+    def checkConstraints(attributes: Seq[Protos.Attribute], constraints: Seq[Constraint]): Boolean = {
+      constraints.foreach { c =>
+        if (!c.matches(attributes)) {
+          return false
+        }
+      }
+
```
elingg: Why is there an extra space here

elingg:
Just a couple of nits. LGTM

wndhydrnt:
Fixed the issues.

elingg:
:+1:  Any objections, @brndnmtthws?

brndnmtthws:
Looks good to me. One day I'd like to see Marathon and Chronos sharing the same constraints code.

kangman:
Hi when will this be available in a stable release?

elingg:
Since we've gotten a number of changes and bug fixes in, we will be releasing a new version of Chronos (2.3.4) in the next week or two.

kangman:
Thanks @elingg  just fyi the documentation gave me some false hope that we already had this feature.  Here is the [entry](https://mesos.github.io/chronos/docs/api.html#sample-job).

elingg:
You're in luck. The release should be out sooner than expected. We're looking at this week.

kangman:
Thanks @elingg !!  Made my day~

elingg:
:+1:

PR #432, opened 2015-04-21
Add try-catch block for initialization

depay:
Chronos may be hanging when exception happens during initing. Althought
it's hard to reappear, it should be handled.


elingg:
Looks good, was this well tested?

depay:
Tested. @elingg

elingg:
:+1:

Issue #394, opened 2015-03-19
Better logging for insufficient resources


juicedM3:
In the file src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala on line 129:

<pre>
log.warning("Insufficient resources remaining for task '%s', will append to queue\n.".format(taskId))
</pre>

It would help to know if the insufficient resource is CPU, mem or disk.  I had an issue where it ended up being disk.  Being new to Mesos & Chronos I ended up banging my head for a couple of days before finally adding some debug statements to the source code and recompiling.

- Mesos 0.21.1 (w/ single slave for testing) 
- ZooKeeper 3.4.6
- Marathon 0.8.0
- Docker 1.0.1
- Chronos 2.3.2
- Ubuntu 14.04.

I compiled Mesos from source and for whatever reason, the slave ended up with it's disk resource set to 188MB (what?!?). A simple "Hello World" Chronos example ended up failing because it was asking for 256MB of disk. Surprisingly, all other simple tests (Mesos' examples, Docker and Marathon) worked for me.

elingg:
If you have time to submit PR with better logging, that would be great. Thanks for reporting this issue.

PR #434, opened 2015-04-24
Better logging for insufficient resources

bitantics:
Add better logging for when a job cannot be scheduled due to insufficient resources.

Resolution for https://github.com/mesos/chronos/issues/394

@elingg 
@michaeljin


elingg:
:+1:

PR #435, opened 2015-04-27
FIX: Handle None return in getTask

mkhettry:
Fix error on chronos startup. 

This fixes a bug I encountered recently. I had deleted some tasks from the chronos UI and when I bounced chronos, it wouldn't come up. The error in the logs pointed to the culprit here:

```
java.util.NoSuchElementException: None.get
```

It looks like code exists to handle the case when jobOption is empty, but we still unconditionally dereference the jobOption.


elingg:
:+1:

PR #436, opened 2015-04-28
Introduce JobsObserver interface and place both JobStats and error notification behind that interface

azakkerman:
I ran all tests and they seem to pass.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
+++ b/src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
@@ -8,11 +8,11 @@ import org.apache.mesos.chronos.scheduler.jobs._
 import org.joda.time.{DateTimeZone, DateTime}
 
 class JobNotificationObserver @Inject()(val notificationClients: List[ActorRef] = List(),
-                                      val clusterName: Option[String] = None) extends JobsObserver {
+                                      val clusterName: Option[String] = None) {
   private[this] val log = Logger.getLogger(getClass.getName)
   val clusterPrefix = clusterName.map(name => s"[$name]").getOrElse("")
 
-  override def onEvent(event: JobEvent): Unit = event match {
+  def asObserver: JobsObserver.Observer = {
     case JobRemoved(job) => sendNotification(job, "%s [Chronos] Your job '%s' was deleted!".format(clusterPrefix, job.name), None)
     case JobDisabled(job, cause) => sendNotification(
       job,
@@ -24,8 +24,6 @@ class JobNotificationObserver @Inject()(val notificationClients: List[ActorRef] 
         .format(DateTime.now(DateTimeZone.UTC), job.retries, taskStatus.getTaskId.getValue)
       sendNotification(job, "%s [Chronos] job '%s' failed!".format(clusterPrefix, job.name),
         Some(TaskUtils.appendSchedulerMessage(msg, taskStatus)))
-
-    case _ =>
   }
 
   def sendNotification(job: BaseJob, subject: String, message: Option[String] = None) {
--- a/src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
+++ b/src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
@@ -7,7 +7,7 @@ import com.google.inject.Inject
 import org.apache.mesos.chronos.scheduler.jobs._
 import org.joda.time.{DateTimeZone, DateTime}
 
-class NotifyingJobsObserver @Inject()(val notificationClients: List[ActorRef] = List(),
+class JobNotificationObserver @Inject()(val notificationClients: List[ActorRef] = List(),
                                       val clusterName: Option[String] = None) extends JobsObserver {
   private[this] val log = Logger.getLogger(getClass.getName)
   val clusterPrefix = clusterName.map(name => s"[$name]").getOrElse("")
--- a/src/main/scala/org/apache/mesos/chronos/notification/NotifyingJobsObserver.scala
+++ b/src/main/scala/org/apache/mesos/chronos/notification/NotifyingJobsObserver.scala
@@ -13,6 +13,7 @@ class NotifyingJobsObserver @Inject()(val notificationClients: List[ActorRef] = 
   val clusterPrefix = clusterName.map(name => s"[$name]").getOrElse("")
 
   override def onEvent(event: JobEvent): Unit = event match {
+    case JobRemoved(job) => sendNotification(job, "%s [Chronos] Your job '%s' was deleted!".format(clusterPrefix, job.name), None)
     case JobDisabled(job, cause) => sendNotification(
       job,
       "%s [Chronos] job '%s' disabled".format(clusterPrefix, job.name),
--- a/src/main/scala/org/apache/mesos/chronos/notification/NotifyingJobsObserver.scala
+++ b/src/main/scala/org/apache/mesos/chronos/notification/NotifyingJobsObserver.scala
@@ -1,0 +1,42 @@
+package org.apache.mesos.chronos.notification
+
+import java.util.logging.Logger
+
+import akka.actor.ActorRef
+import com.google.inject.Inject
+import org.apache.mesos.chronos.scheduler.jobs._
+import org.joda.time.{DateTimeZone, DateTime}
+
+class NotifyingJobsObserver @Inject()(val notificationClients: List[ActorRef] = List(),
+                                      val clusterName: Option[String] = None) extends JobsObserver {
+  private[this] val log = Logger.getLogger(getClass.getName)
+  val clusterPrefix = clusterName.map(name => s"[$name]").getOrElse("")
+
+  override def onEvent(event: JobEvent): Unit = event match {
+    case JobDisabled(job, cause) => sendNotification(
+      job,
+      "%s [Chronos] job '%s' disabled".format(clusterPrefix, job.name),
+      Some(cause))
+
+    case JobRetriesExhausted(job, taskStatus, attempts) =>
+      val msg = "\n'%s'. Retries attempted: %d.\nTask id: %s\n"
+        .format(DateTime.now(DateTimeZone.UTC), job.retries, taskStatus.getTaskId.getValue)
+      sendNotification(job, "%s [Chronos] job '%s' failed!".format(clusterPrefix, job.name),
+        Some(TaskUtils.appendSchedulerMessage(msg, taskStatus)))
+
+    case _ =>
+  }
+
+  def sendNotification(job: BaseJob, subject: String, message: Option[String] = None) {
+    for (client <- notificationClients) {
+      val subowners = job.owner.split("\\s*,\\s*")
+      for (subowner <- subowners) {
+        log.info("Sending mail notification to:%s for job %s using client: %s".format(subowner, job.name, client))
+        client !(job, subowner, subject, message)
+      }
+    }
+
+    log.info(subject)
+  }
+
+}
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
@@ -17,6 +17,7 @@ import com.fasterxml.jackson.databind.module.SimpleModule
 
 import com.codahale.metrics.annotation.Timed
 import com.google.inject.Inject
+import org.apache.mesos.chronos.scheduler.jobs.stats.JobStats
 import org.joda.time.{DateTime, DateTimeZone}
 
 import scala.collection.mutable.ListBuffer
@@ -33,6 +34,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
                                       val jobGraph: JobGraph,
                                       val configuration: SchedulerConfiguration,
                                       val cassandraConfig: CassandraConfiguration,
+                                      val jobStats: JobStats,
                                       val jobMetrics: JobMetrics) {
 
   private[this] val log = Logger.getLogger(getClass.getName)
@@ -100,8 +102,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
             }
         }
       }
-      jobScheduler.sendNotification(job, "[Chronos] - Your job '%s' was deleted!".format(jobName))
-
+      // No need to send notifications here, the jobScheduler.deregisterJob will do it
       jobScheduler.deregisterJob(job, persist = true)
       Response.noContent().build
     } catch {
@@ -123,7 +124,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
       require(!jobOpt.isEmpty, "Job '%s' not found".format(jobName))
 
       val histoStats = jobMetrics.getJobHistogramStats(jobName)
-      val jobStatsList: List[TaskStat] = jobScheduler.jobStats.getMostRecentTaskStatsByJob(jobOpt.get, cassandraConfig.jobHistoryLimit())
+      val jobStatsList: List[TaskStat] = jobStats.getMostRecentTaskStatsByJob(jobOpt.get, cassandraConfig.jobHistoryLimit())
       val jobStatsWrapper = new JobStatWrapper(jobStatsList, histoStats)
 
       val wrapperStr = objectMapper.writeValueAsString(jobStatsWrapper)
@@ -183,7 +184,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
           require(num >= 0,
             "numAdditionalElementsProcessed (%d) is not positive".format(num))
 
-          jobScheduler.jobStats.updateTaskProgress(jobOpt.get, taskId, num)
+          jobStats.updateTaskProgress(jobOpt.get, taskId, num)
         }
         case None =>
       }
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
@@ -91,7 +91,7 @@ class MainModule(val config: SchedulerConfiguration with HttpConf)
                             mesosSchedulerDriver: MesosDriverFactory,
                             curator: CuratorFramework,
                             leaderLatch: LeaderLatch,
-                            jobsObserver: JobsObserver,
+                            jobsObserver: JobsObserver.Observer,
                             metrics: JobMetrics): JobScheduler = {
     new JobScheduler(
       scheduleHorizon = Seconds.seconds(config.scheduleHorizonSeconds()).toPeriod,
@@ -110,9 +110,9 @@ class MainModule(val config: SchedulerConfiguration with HttpConf)
 
   @Singleton
   @Provides
-  def provideJobsObservers(jobStats: JobStats, notificationClients: List[ActorRef]): JobsObserver = {
+  def provideJobsObservers(jobStats: JobStats, notificationClients: List[ActorRef]): JobsObserver.Observer = {
     val notifier = new JobNotificationObserver(notificationClients, config.clusterName.get)
-    JobsObserver.composite(List(notifier, jobStats))
+    JobsObserver.composite(List(notifier.asObserver, jobStats.asObserver))
   }
 
   @Singleton
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
@@ -110,9 +110,9 @@ class MainModule(val config: SchedulerConfiguration with HttpConf)
 
   @Singleton
   @Provides
-  def provideJobsObservers(stats: JobStats, notificationClients: List[ActorRef]): JobsObserver = {
+  def provideJobsObservers(jobStats: JobStats, notificationClients: List[ActorRef]): JobsObserver = {
     val notifier = new NotifyingJobsObserver(notificationClients, config.clusterName.get)
-    JobsObserver.composite(List(notifier, stats))
+    JobsObserver.composite(List(notifier, jobStats))
   }
 
   @Singleton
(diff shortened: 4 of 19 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -100,8 +102,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
             }
         }
       }
```
elingg: This deleted job notification will no longer be sent. No notification is sent in deregisterJob. We should still send the notification here.

src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
```diff
@@ -102,12 +102,17 @@ class MainModule(val config: SchedulerConfiguration with HttpConf)
...
-      notificationClients = notificationClients,
+      jobsObserver = jobsObserver,
       failureRetryDelay = config.failureRetryDelayMs(),
       disableAfterFailures = config.disableAfterFailures(),
-      jobMetrics = metrics,
-      jobStats = stats,
-      clusterName = config.clusterName.get)
+      jobMetrics = metrics)
+  }
+
```
elingg: Where is this method (provideJobsObservers) called?

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
azakkerman: deregisterJob calls jobsObserver.onEvent(JobRemoved(job)), I just need to handle that in NotifyingJobObserver

src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
azakkerman: it is Guice @Provides annotated factory method, if I understand Guice correctly. So whenever JobsObserver instance is needed a singleton instance is provided. https://github.com/google/guice/wiki/ProvidesMethods

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
elingg: sounds good, could you add that so that this notification is still sent?

elingg: Ah, I see it was just added. Thanks!

src/main/scala/org/apache/mesos/chronos/notification/NotifyingJobsObserver.scala
```diff
@@ -0,0 +1,43 @@
+package org.apache.mesos.chronos.notification
+
+import java.util.logging.Logger
+
+import akka.actor.ActorRef
+import com.google.inject.Inject
+import org.apache.mesos.chronos.scheduler.jobs._
+import org.joda.time.{DateTimeZone, DateTime}
+
```
elingg: I would prefer the naming JobNotificationObserver

src/main/scala/org/apache/mesos/chronos/scheduler/config/MainModule.scala
elingg: Makes sense, thanks

src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import akka.actor.ActorRef
+import com.google.inject.Inject
+import org.apache.mesos.chronos.scheduler.jobs._
+import org.joda.time.{DateTimeZone, DateTime}
+
+class JobNotificationObserver @Inject()(val notificationClients: List[ActorRef] = List(),
+                                      val clusterName: Option[String] = None) extends JobsObserver {
+  private[this] val log = Logger.getLogger(getClass.getName)
+  val clusterPrefix = clusterName.map(name => s"[$name]").getOrElse("")
```
elingg: Let's evaluate clusterPrefix inside JobRetriesExhausted (only where it is used)

azakkerman: It is also used in JobRemoved case

elingg: ok, makes sense

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
```diff
@@ -429,37 +400,17 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
...
-                case Some(msg) =>
-                  sendNotification(job, "%s [Chronos] JOB DISABLED: '%s'".format(clusterPrefix, job.name),
-                    Some("\nFailed at '%s', %d failures since last success\nTask id: %s\nThe scheduler provided this message:\n\n%s"
-                      .format(DateTime.now(DateTimeZone.UTC), newJob.errorsSinceLastSuccess,
-                        taskId, msg)))
-                case None =>
-                  sendNotification(job, "%s [Chronos] JOB DISABLED: '%s'".format(clusterPrefix, job.name),
-                    Some("\nFailed at '%s', %d failures since last success\nTask id: %s\n"
-                      .format(DateTime.now(DateTimeZone.UTC), newJob.errorsSinceLastSuccess, taskId)))
-              }
```
elingg: We are ignoring taskStatus.getMessage here it seems. We should not remove that logging.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
```diff
@@ -429,37 +400,17 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
...
-                      .format(DateTime.now(DateTimeZone.UTC), job.retries,
-                        taskId, msg)))
-                case None =>
-                  sendNotification(job, "%s [Chronos] job '%s' failed!".format(clusterPrefix, job.name),
-                    Some("\n'%s'. Retries attempted: %d.\nTask id: %s\n"
-                      .format(DateTime.now(DateTimeZone.UTC), job.retries, taskId)))
-              }
+              jobsObserver.onEvent(JobRetriesExhausted(job, taskStatus, attempt))
             }
             jobMetrics.updateJobStatus(jobName, success = false)
```
elingg: We are ignoring taskStatus.getMessage here it seems. We should not remove that logging.

azakkerman: Both here and above it is taken care of in JobNotificationObserver, see calls to TaskUtils.appendSchedulerMessage(msg, taskStatus), and see the code for appendSchedulerMessage() utility method

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
```diff
@@ -32,6 +32,15 @@ object TaskUtils {
     taskIdPattern.findFirstIn(taskIdString).nonEmpty
   }
```
elingg: RichBoolean is only used in appendSchedulerMessage and is a bit confusing in terms of readability. Could we instead just do the if (taskStatus.hasMessage && taskStatus.getMessage.nonEmpty) Some(taskStatus.getMessage) else None inside the appendSchedulerMessage?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
```diff
@@ -360,7 +353,16 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
   }
 
-  def jobQueued(job: BaseJob, attempt: Int) {
+  override def onEvent(event: JobEvent): Unit = event match {
+    case JobRemoved(job) => removeJobState(job)
+    case JobQueued(job, taskId, attempt) => jobQueued(job, taskId, attempt)
+    case JobStarted(job, taskStatus, attempt) => jobStarted(job, taskStatus, attempt)
+    case JobFinished(job, taskStatus, attempt) => jobFinished(job, taskStatus, attempt)
+    case JobFailed(job, taskStatus, attempt) => jobFailed(job, taskStatus, attempt)
+    case _ => // do nothing
```
elingg: Should we throw an exception or at least log an error for case _ ?

src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
```diff
@@ -0,0 +1,43 @@
...
+      job,
+      "%s [Chronos] job '%s' disabled".format(clusterPrefix, job.name),
+      Some(cause))
+
+    case JobRetriesExhausted(job, taskStatus, attempts) =>
+      val msg = "\n'%s'. Retries attempted: %d.\nTask id: %s\n"
+        .format(DateTime.now(DateTimeZone.UTC), job.retries, taskStatus.getTaskId.getValue)
+      sendNotification(job, "%s [Chronos] job '%s' failed!".format(clusterPrefix, job.name),
+        Some(TaskUtils.appendSchedulerMessage(msg, taskStatus)))
+
```
elingg: Should we throw an exception or at least log an error for case _ ?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
```diff
@@ -85,7 +79,6 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
    * Queries Cassandra table for past and current job statistics by jobName
    * and limits by numTasks. The result is not sorted by execution time
```
elingg: Thanks for fixing the ScalaDoc!

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
azakkerman: I would have preferred to use Scalaz's method on Booleans (see option[A](a: => A) in http://docs.typelevel.org/api/scalaz/nightly/index.html#scalaz.syntax.std.BooleanOps)

and encourage more functional style. I am pretty sure there are more cases where it can be used, but if you prefer the more verbose style of
if (condition)
   Some(value)
else
   None

It is up to you.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
azakkerman: I don't think so, JobStats is not (and never was) interested in the other events. If you want to add some kind of info or debug logging, that would be ok

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
```diff
@@ -489,54 +491,30 @@ class JobStats @Inject()(clusterBuilder: Option[Cluster.Builder], config: Cassan
...
-            jobName=Some(job.name),
-            jobOwner=Some(job.owner),
-            jobSchedule=jobSchedule,
-            jobParents=jobParents,
-            taskState=Some(taskStatus.getState.toString),
-            slaveId=Some(taskStatus.getSlaveId.getValue),
-            message=Some(taskStatus.getMessage),
-            attempt=Some(attempt),
-            isFailure=Some(true))
-  }
```
elingg: can we have a more descriptive variable name other than j?

src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
azakkerman: The notification observer is not currently interested in other events, throwing exception here would unwind the stack for no good reason. In general, these observers are probably better written as PartialFunctions. I could do that refactoring.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
azakkerman: np, IntelliJ was complaining so I wanted to quiet it down :)

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
azakkerman: jobOrName?

src/test/scala/org/apache/mesos/chronos/scheduler/jobs/MockUtils.scala
```diff
@@ -0,0 +1,17 @@
+package org.apache.mesos.chronos.scheduler.jobs
```
elingg: Could we name this class MockJobUtils?

elingg:
Great work, @azakkerman! I think use of an common interface for JobStats and JobNotifications is much cleaner. Thanks for addressing my comments.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
elingg: Ah, thanks for clarifying. That utility method does take care of it.

src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
elingg: PartialFunctions sounds good or at least there should be a log info message letting us know that the event is not currently handled.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
elingg: PartialFunctions sounds good or at least there should be a log info message letting us know that the event is not currently handled.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
elingg: sure, another option is jobInfo or jobNameOrJob.

src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
azakkerman: Would a generic log.info() for any observer that does not handle a specific case suffice?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
elingg: We might want to add a helper method to TaskUtils with the appropriate naming if we see this behavior repeatedly. Since this is only in 1 method, I think for readability sake, this makes more sense.

src/main/scala/org/apache/mesos/chronos/notification/JobNotificationObserver.scala
elingg: Yes, sounds good. Then we should have full coverage.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
azakkerman: I see that scalaz is one of external libs, can I just use the scalaz version of this method?

elingg: sure, even better.

azakkerman: couldn't get it to work, probably because scalaz is not a direct dependency in the maven project, not sure. Will simply keep as if/else

elingg:
LGTM. @brndnmtthws will also be reviewing this since this is a significant PR.

src/test/scala/org/apache/mesos/chronos/scheduler/jobs/MockJobUtils.scala
```diff
@@ -0,0 +1,21 @@
+package org.apache.mesos.chronos.scheduler.jobs
+
+import org.apache.mesos.chronos.scheduler.graph.JobGraph
+import org.apache.mesos.chronos.scheduler.state.PersistenceStore
+import org.joda.time.Period
+import org.specs2.mock._
+
+object MockJobUtils extends Mockito {
```
brndnmtthws: Thanks for adding the test!

brndnmtthws:
This looks great to me. Before we merge, can you please test this in a cluster? I think we can use the test-suite for that purpose. @elingg can you help with that?

elingg:
@brndnmtthws, yes, I will test in my own cluster. One thing is that we haven't had a chance to hook up  Chronos with Cassandra in the test cluster, and I'm not sure we will get to it today. Do you think we should get that up before merging? 

@azakkerman, could you also try testing in a cluster as well? @kensipe should be able to help you with set up as well.

brndnmtthws:
It's pretty easy, just point it at C* (using Mesos-DNS names). You'll have to create the keyspace manually, but we should really have it happen within Chronos (`CREATE KEYSPACE IF NOT EXISTS [derp] WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 3 };`).

elingg:
ok @brndnmtthws I will be sure test with Cassandra before merging it. If @azakkerman also has a chance to test earlier than I do, that would be great.

azakkerman:
I am trying to bring up DCOS in AWS with Ken's help to try testing it.

azakkerman:
I have a DCOS cluster running in AWS with Cassandra deployed, but don't quite know how to configure chronos to run against that Cassandra instance. Who is a good resource for this?

elingg:
Hello @azakkerman, you will have to run your version on Chronos with the cassandra options enabled and simply point to cassandra-dcos-node.cassandra.dcos.mesos for the cassandra hostname. You will also need to create the keyspace as Brenden mentioned above:
CREATE KEYSPACE IF NOT EXISTS [derp] WITH REPLICATION = { 'class' : 'SimpleStrategy', 'replication_factor' : 3 };)

elingg:
If you would like to discuss over chat since I am WFH today, maybe you could ask @kensipe to add you to our slack channel. We can talk via IM. I will also be in the office tomorrow.

elingg:
Here are the Cassandra options for Chronos:

"cassandra_consistency”: Consistency to use for Cassandra (default = ANY).

“cassandra_contact_points”: Comma separated list of contact points for Cassandra.

“cassandra_keyspace”: Keyspace to use for Cassandra (default = metrics).

“cassandra_port”: Port for Cassandra (default = 9042).

“cassandra_table”: Table to use for Cassandra (default = chronos).

“cassandra_ttl”: TTL for records written to Cassandra (default = 31536000).

azakkerman:
I was able to launch chronos with my custom jar, and write to cassandra. I see rows in cassandra corresponding to my test job instances. What sort of tests should I perform?

azakkerman:
cqlsh> select * from metrics.chronos
   ... ;

 id                       | ts                       | attempt | is_failure | job_name | job_owner          | job_parents | job_schedule                     | message | slave_id                                | task_state
--------------------------+--------------------------+---------+------------+----------+--------------------+-------------+----------------------------------+---------+-----------------------------------------+---------------
 ct:1430339602000:0:foo1: | 2015-04-29 20:33:22+0000 |       0 |       null |     foo1 | akkerman@gmail.com |        null | R1/2015-04-29T20:35:22.000Z/PT1M |    null | 20150429-173600-1157627914-5050-1074-S3 |  TASK_RUNNING
 ct:1430339602000:0:foo1: | 2015-04-29 20:33:22+0000 |       0 |       null |     foo1 | akkerman@gmail.com |        null | R1/2015-04-29T20:35:22.000Z/PT1M |    null | 20150429-173600-1157627914-5050-1074-S3 | TASK_FINISHED
 ct:1430339662000:0:foo1: | 2015-04-29 20:34:22+0000 |       0 |       null |     foo1 | akkerman@gmail.com |        null | R0/2015-04-29T20:36:22.000Z/PT1M |    null | 20150429-173600-1157627914-5050-1074-S1 |  TASK_RUNNING
 ct:1430339662000:0:foo1: | 2015-04-29 20:34:22+0000 |       0 |       null |     foo1 | akkerman@gmail.com |        null | R0/2015-04-29T20:36:22.000Z/PT1M |    null | 20150429-173600-1157627914-5050-1074-S1 | TASK_FINISHED
 ct:1430339542265:0:foo1: | 2015-04-29 20:32:22+0000 |       0 |       null |     foo1 | akkerman@gmail.com |        null | R0/2015-04-29T20:33:22.000Z/PT1M |    null | 20150429-173600-1157627914-5050-1074-S4 |  TASK_RUNNING
 ct:1430339542265:0:foo1: | 2015-04-29 20:32:26+0000 |       0 |       null |     foo1 | akkerman@gmail.com |        null | R0/2015-04-29T20:33:22.000Z/PT1M |    null | 20150429-173600-1157627914-5050-1074-S4 | TASK_FINISHED

elingg:
Nice job! Can you see the job history in the UI?

azakkerman:
screen shot 2015-04-29 at 2 00 29 pm

elingg:
Looking good! Can you try a job that takes longer to run to confirm you can see some stats? Here is an example of what the UI should look like: https://github.com/mesos/chronos/pull/402

azakkerman:
I ran a sleep job that exits 1 in the end and verified that the email notification still works. Also deleted a job and got some emails. What specific stats are you looking for?

elingg:
Notification emails are also a great test. If you run a job that runs longer than 0 or 1 sec (let's say sleep 5), you should see stats like in the first screen shot in this link:  https://github.com/mesos/chronos/pull/402

elingg:
After that, we should have tested sufficiently!

azakkerman:
Btw, it looks like if one had a job with limited repeat count and the job gets disabled (because repeat count is now 0), there is no way to re-enable the job by incrementing the repeat count and updating start time, the job remains disabled. I think this is a bug in the UI.
screen shot 2015-04-29 at 2 45 38 pm

elingg:
Very possible that it is a UI bug. Anyway your test cases sound great! I think we are good to merge. Any final objections @brndnmtthws?

brndnmtthws:
Wonderful!

PR #437, opened 2015-04-29
Cleanup of compiler warnings and complete JobStats refactoring

azakkerman:
This PR contains #436 in it. It adds proper firing of JobQueued via the new interface and adds JobExpired callback. It also makes corresponding methods in JobStats private, to make sure that they are not called from anywhere but the Observer interface.

Out of boredom, I cleaned up a good number of warnings in several classes to make IntelliJ happy.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
@@ -103,13 +103,15 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
       jobScheduler.deregisterJob(job, persist = true)
       Response.noContent().build
     } catch {
-      case ex: IllegalArgumentException =>
+      case ex: IllegalArgumentException => {
         log.log(Level.INFO, "Bad Request", ex)
         Response.status(Response.Status.BAD_REQUEST).entity(ex.getMessage)
           .build()
-      case ex: Exception =>
+      }
+      case ex: Exception => {
         log.log(Level.WARNING, "Exception while serving request", ex)
         Response.serverError().build
+      }
     }
   }
 
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
@@ -9,9 +9,6 @@ import org.apache.mesos.chronos.scheduler.config.{CassandraConfiguration, Schedu
 import org.apache.mesos.chronos.scheduler.graph.JobGraph
 import org.apache.mesos.chronos.scheduler.jobs._
 
-import com.codahale.metrics.Histogram
-import com.datastax.driver.core.Row
-import com.datastax.driver.core.ColumnDefinitions
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.module.SimpleModule
 
@@ -121,7 +118,7 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
   def getStat(@PathParam("jobName") jobName: String): Response = {
     try {
       val jobOpt = jobGraph.lookupVertex(jobName)
-      require(!jobOpt.isEmpty, "Job '%s' not found".format(jobName))
+      require(jobOpt.nonEmpty, "Job '%s' not found".format(jobName))
 
       val histoStats = jobMetrics.getJobHistogramStats(jobName)
       val jobStatsList: List[TaskStat] = jobStats.getMostRecentTaskStatsByJob(jobOpt.get, cassandraConfig.jobHistoryLimit())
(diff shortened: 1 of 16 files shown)


elingg:
Nice work, @azakkerman! I'll take a look at this once the https://github.com/mesos/chronos/pull/436 is merged. Could you make sure this compiles (I see a compile error) and also that this is also tested.

azakkerman:
Travis CI seems not to be picking up the right versions of the files for build, if you look at the line of code where it is failing and the actual version 9e3f2c04543303d53f0e2e43d79238f7fac2f9d0 of TaskManagerSpec, you will clearly see it. You can check out my version of the code and it will build just fine. I am at a loss here...

elingg:
Strange. Once we merge the other PR, you can try rebasing with master.

elingg:
Another thought is did you try mvn clean package locally? Sometimes without the ```clean```, it does not pick up the latest artifacts.

azakkerman:
mvn clean package in my git repo gives:

[INFO]
------------------------------------------------------------------------

[INFO] BUILD SUCCESS

[INFO]
------------------------------------------------------------------------

[INFO] Total time: 04:12 min

[INFO] Finished at: 2015-04-29T17:06:15-07:00

[INFO] Final Memory: 51M/590M

[INFO]
------------------------------------------------------------------------

elingg:
Hm, strange. I will take a look myself once the other PR is merged and this is rebased with master.

brndnmtthws:
As per @elingg's comment, this needs a rebase.

azakkerman:
I rebased, fixed compile and pushed. This time Travis is failing for some unrelated reasons, it seems. Can you take a look?

azakkerman:
Alright, I forced a rebuild by closing and reopening the PR, everything is good now. Please, review.

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -173,32 +170,27 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
...
           //NOTE: 0 is a valid value
           require(num >= 0,
             "numAdditionalElementsProcessed (%d) is not positive".format(num))
 
           jobStats.updateTaskProgress(jobOpt.get, taskId, num)
-        }
-        case None =>
       }
       Response.noContent().build
     } catch {
```
elingg: We need curly braces here for multiple lines in the case class. Also, we want to make sure that there is no logical change here. See the scala style guide: http://docs.scala-lang.org/style/control-structures.html/ case - Omit braces if the case expression fits on a single line. Otherwise, use curly braces for clarity (even though they are not required by the parser).

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobStatWrapperSerializer.scala
```diff
@@ -5,8 +5,8 @@ import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.databind.JsonSerializer
 import com.fasterxml.jackson.databind.SerializerProvider
 
-import org.joda.time.{DateTime, Duration}
-import org.joda.time.format.{DateTimeFormat, DateTimeFormatter, PeriodFormat, PeriodFormatter, PeriodFormatterBuilder}
+import org.joda.time.DateTime
+import org.joda.time.format.{DateTimeFormat, PeriodFormatterBuilder}
```
elingg: Very nice clean up of this class!

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
```diff
@@ -507,21 +507,24 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
...
             //TODO(FL): Rethink passing the dispatch queue all the way down to the ScheduledTask.
             val task = new ScheduledTask(TaskUtils.getTaskId(job, nextDate), nextDate, job, taskManager)
-            return (Some(task), stream.tail())
+            return (Some(task), stream.tail)
           }
-          //The nextDate has passed already beyond epsilon.
-          //TODO(FL): Think about the semantics here and see if it always makes sense to skip ahead of missed schedules.
+          // Next instance is too far in the future
+          // Needs to be scheduled at a later time, after schedule horizon.
           if (!nextDate.isBefore(now)) {
```
elingg: We still need the original logic here jobsObserver.apply(JobSkipped(job, nextDate)) in order to notify the observer that we have skipped the job.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
```diff
@@ -507,21 +507,24 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
...
-            jobsObserver.apply(JobSkipped(job, nextDate))
             return (None, Some(stream))
           }
-          //Needs to be scheduled at a later time, after schedule horizon.
+          // Next instance is too far in the past (beyond epsilon)
+          //TODO(FL): Think about the semantics here and see if it always makes sense to skip ahead of missed schedules.
           log.fine("No need to work on schedule: '%s' yet".format(nextDate))
```
elingg: We do not want to add  jobsObserver.apply(JobSkipped(job, nextDate)) here, because we are running the job now. We don't want to notify the observer that the job was skipped when it's about to get executed.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/stats/JobStats.scala
```diff
@@ -13,7 +13,6 @@ import org.apache.mesos.chronos.scheduler.jobs._
 import org.joda.time.DateTime
 
 import scala.collection.JavaConverters._
```
elingg: We should import HashMap here, but Map is not necessary

elingg:
Thanks for the clean up @azakkerman! I made some comments above. Could you address them? Also, were you able to test this?

@brndnmtthws, this is a fairly large clean up PR. Could you also take a look to be safe?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
azakkerman: It was me erroneously adding this call before. It does not belong here, we are not skipping any instances in this branch of the code, this branch is when the next instance to be scheduled is in the future.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
azakkerman: If you follow the logic, the first check on line 512 checks if he next instance is in the scheduling window, if so, return it.
The next check was to see if the remaining instances are too far in the future, if so, nothing to add to the queue, but nothing is skipped either.
We get here IFF the instance we are looking at is from the past, from _before_ the scheduling window. When that happens, notice that we just drop the head and make a recursive call via next with remaining scheduled times.

So, I argue that this call to skipped is now in the correct place, while before it was not

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
azakkerman: Are you referring to lines below the comment in exception matching case?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
azakkerman: And the reason I noticed was because I ran chronos and saw a bunch of unexpected skips in the log, then I figured that the skipped callback is being invoked in a wrong place.

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
elingg: correct, the lines below

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
elingg: Fiar enough

elingg:
LGTM, unless there are any final objections from @brndnmtthws, let's merge.

PR #439, opened 2015-05-04
Fix dependent job not run if it got updated

wndhydrnt:
This fixes an issue where a job that depends on two or more parent jobs
is not executed if it was updated while some of the parent jobs
have finished while others are still running.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/DependentJobResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/DependentJobResource.scala
@@ -57,10 +57,16 @@ class DependentJobResource @Inject()(
         val parents = jobGraph.parentJobs(newJob)
         oldJob match {
           case j: DependencyBasedJob =>
-            val oldParents = jobGraph.parentJobs(j)
-            oldParents.map(x => jobGraph.removeDependency(x.name, oldJob.name))
+            val newParentNames = parents.map(_.name)
+            val oldParentNames = jobGraph.parentJobs(j).map(_.name)
+
+            if (newParentNames != oldParentNames) {
+              oldParentNames.foreach(jobGraph.removeDependency(_, oldJob.name))
+              newParentNames.foreach(jobGraph.addDependency(_, newJob.name))
+            }
             jobScheduler.removeSchedule(j)
           case j: ScheduleBasedJob =>
+            parents.foreach(p => jobGraph.addDependency(p.name, newJob.name))
         }
 
         jobScheduler.updateJob(oldJob, newJob)
@@ -71,7 +77,6 @@ class DependentJobResource @Inject()(
           new String(JobUtils.toBytes(oldJob), Charsets.UTF_8),
           new String(JobUtils.toBytes(newJob), Charsets.UTF_8)))
 
-        parents.map(x => jobGraph.addDependency(x.name, newJob.name))
         Response.noContent().build()
       }
     } catch {


wndhydrnt:
Some more background information.

Given the following dependency graph:
chronos_dep
* `parent` runs for five seconds. It's just there to trigger the other jobs.
* `layer1-job1` takes 60 seconds to complete
* `layer1-job2` takes 30 seconds to complete
* `layer2-job1` is just there to be triggered by its parent jobs

Without this fix, this happens:

1. `parent` starts
2. `parent` finishes and triggers `layer1-job1` and `layer1-job2`
3. `layer1-job2` finishes but `layer1-job1` is still running
4. A user updates the configuration of `layer2-job1`
5. `layer1-job1` finishes
6. `layer2-job1` is not executed

This happens because [DependentJobResource.handleRequest](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/api/DependentJobResource.scala#L32) first calls [JobGraph.removeDependency](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala#L101) and then [JobGraph.addDependency](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala#L92). Doing so resets the value of `edgeInvocationCount ` for the edge between `layer1-job2` and `layer2-job1` to `0` which leads to `count` and `min` not being equal [here](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala#L163).

I would expect that the job is executed if the parents of a job have not changed during an update and that is what this PR fixes.

src/main/scala/org/apache/mesos/chronos/scheduler/api/DependentJobResource.scala
```diff
@@ -57,10 +57,16 @@ class DependentJobResource @Inject()(
...
-            oldParents.map(x => jobGraph.removeDependency(x.name, oldJob.name))
+            val newParentNames = parents.map(_.name)
+            val oldParentNames = jobGraph.parentJobs(j).map(_.name)
+
+            if (newParentNames != oldParentNames) {
+              oldParentNames.foreach(jobGraph.removeDependency(_, oldJob.name))
+              newParentNames.foreach(jobGraph.addDependency(_, newJob.name))
+            }
             jobScheduler.removeSchedule(j)
           case j: ScheduleBasedJob =>
```
elingg: Question, don't we need to remove the dependency to the old jobs here?

wndhydrnt: No, that is not necessary, because a former `ScheduleBasedJob` has no parents/dependencies.
Also it is not possible as `JobGraph.parentJobs` only accepts values of type `DependencyBasedJob`.

elingg: This makes sense, I think this line: ``parents.foreach(p => jobGraph.addDependency(p.name, newJob.name))`` needs to be outside the case statement as it was before, though. This is because both types of jobs can have child jobs. DependencyBasedJobs will be the only one with parents.

wndhydrnt: If this is done then the bug is reintroduced because `JobGraph.addDependency` [resets](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/graph/JobGraph.scala#L97) the counter of `edgeInvocationCount`.
Adding dependencies is executed for both type of jobs.
`newParentNames.foreach(jobGraph.addDependency(_, newJob.name))` does it for a `DependencyBasedJob` and

```
case j: ScheduleBasedJob =>
  parents.foreach(p => jobGraph.addDependency(p.name, newJob.name))
```
does it for a former-`ScheduleBasedJob`-now-`DependencyBasedJob`.

It's just that, for a `DependencyBasedJob` that is updated and is not converted from a `ScheduleBasedJob`, the update of its dependencies only happens if the dependencies actually changed.

elingg: Ah, understood! Thanks for the explanation, and I did a further review. We are updating parent dependencies only (and this is only the case of dependencybasedjobs and scheduledjobs which have become dependencybasedjobs)  and we don't want to get the counts out of sync when updating the graph for the dependency based job. We should be good to go.

elingg:
:+1:

wndhydrnt:
Thanks for reviewing this so quickly.

Issue #42, opened 2013-05-28
Chronos's assets won't optimize if server has low RAM.


jeid64:
Hey, I didn't really know if this was an issue or not but it would be nice if it was stated somewhere. I was getting Stackoverflow errors on my VM at the optimization part of the installer.sh script on my VM that had a gig of RAM. I was wondering why my Chronos compile seemed to work besides that, but I wouldn't get anything on the webpage. 0 assets would be put into the JAR with low RAM and there isn't a check in the make script to see if the optimization worked or not. Also, I was slowly stepping up the RAM on the VM and slowly more and more assets would be put into the JAR, so it depends on the RAM amount how much succeeds.

PR #442, opened 2015-05-12
Added outline for docs site.

ssk2:
Moved the docs site to use a similar format as Marathon (https://mesosphere.github.io/marathon).  The docs are mostly the same as what was in the README and docs folder before, just merged and cleaned up a little. 

I've also added a small set of instructions on how to install from packages.


docs/README.md
```diff
@@ -0,0 +1,70 @@
+# Chronos Docs and Website
+
```
elingg: I would rename this "Chronos Documentation Website" to be clear

docs/docs/api.md
```diff
@@ -0,0 +1,357 @@
+---
+title: REST API
+---
+
+# REST API
+
+You can communicate with Chronos using a RESTful [JSON][] API over HTTP.
+Chronos nodes usually listen on `port 8080` for API requests.
+All examples in this section assume that you've found a running leader at `chronos-node.airbnb.com:8080`.
+
```
elingg: Not you, but can we brand this as chronos-node.mesosphere.com:4400. Let's change ports from 8080 to 4400 as the standard since 8080 is typically used by Marathon.

docs/docs/api.md
```diff
@@ -0,0 +1,357 @@
...
+When you have multiple Chronos nodes running, only one of them will be elected as the leader.
+The leader is the only node that responds to API requests, but if you attempt to talk to a non-leader your request will automatically be redirected to a leader.
+
+### Listing Jobs
+
+* Endpoint: __/scheduler/jobs__
+* Method: __GET__
+* Example: `curl -L -X GET chronos-node:8080/scheduler/jobs`
+* Response: JSON data
+
```
elingg: Port 4400?

docs/docs/debugging.md
```diff
@@ -0,0 +1,26 @@
...
+    Task with id 'value: TASK_ID **FINISHED**
+
+To find debug logs on the mesos slave, look in `/tmp/mesos/slaves` on the slave instance (unless you've specifically supplied a different log folder for mesos). For example:
+
+    /tmp/mesos/slaves/
+
+In that dir, the current slave run is timestamped so look for the most recent.
+Under that is a list of frameworks; you're interested in the Chronos framework.
+For example:
+
```
elingg: can we make this <timestamp> instead of STAMP

docs/docs/faq.md
```diff
@@ -1,14 +1,37 @@
...
+If you use the cURL command line tool, you can use the `-L` flag and hit any Chronos node and you will get a
+307 REDIRECT to the leader.
+
+
+### <a name="chronos-zookeeper"></a>How does Chronos use ZooKeeper?
+
+Chronos registers itself with [ZooKeeper][ZooKeeper] at the location `/chronos/state`. This value can be changed via the configuration file.
+
+### <a name="chronos-cassandra"></a>How does Chronos use Cassandra?
+
```
elingg: I would say "job history, reporting, and statistics" here

docs/docs/index.md
```diff
@@ -0,0 +1,106 @@
...
+### Install Dependencies
+
+#### Debian Linux:
+    sudo apt-get install autoconf make gcc cpp patch python-dev git libtool default-jdk default-jdk-builddep default-jre gzip libghc-zlib-dev libcurl4-openssl-dev
+
+#### Fedora Linux:
+    sudo yum install autoconf make gcc gcc-c++ patch python-devel git libtool java-1.7.0-openjdk-devel zlib-devel libcurl-devel openssl-devel cyrus-sasl-devel
+
+Make sure you're using Java 7: `sudo alternatives --config java`
+
```
elingg: Let's have the option here of only installing Mesos and not building it. Most users will want to simply install it.

docs/docs/index.md
```diff
@@ -0,0 +1,106 @@
...
+    git clone https://github.com/apache/mesos.git
+    cd mesos/
+    git checkout 
+    export JAVA_HOME=/usr/lib/jvm/java-1.7.0-openjdk-1.7.0.45.x86_64/
+    ./bootstrap
+    ./configure --with-webui --with-included-zookeeper --disable-perftools --enable-frame-pointers
+    make
+    sudo make install
+
```
elingg: Let's have the option here of only installing Chronos and not building it. Most users will want to simply install it using yum install, for example.

docs/docs/api.md
elingg: If we do update the port, we should also update the default configuration. Maybe this should be a separate PR, however. This PR is already pretty big.

docs/docs/faq.md
elingg: Also, we should probably mention that Cassandra is optional.

docs/docs/webui.md
```diff
@@ -19,7 +20,7 @@ There are also two included example YAML files that make use of this configurati
 * [config/local_scheduler.yml](/mesos/chronos/blob/master/config/local_scheduler.yml#L4)
 * [config/local_scheduler_nozk.yml](/mesos/chronos/blob/master/config/local_scheduler_nozk.yml#L7)
 
-## Build Requirements
+### Build Requirements
```
elingg: It seems that the issue is closed (#42) some time ago so I wouldn't mention it.  @mlunoe, did you encounter any out of memory issue when you built the Chronos UI? Just want to double check.

docs/support.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Getting Support for Chronos
+
+For questions and discussions around Chronos, please use the Google Group "chronos-scheduler":
+[Chronos Scheduler Group](https://groups.google.com/forum/#!forum/chronos-scheduler).
+
+Also join us on IRC in #mesos on freenode.
+
+## Reporting Bugs
+
```
elingg: Nice bug reporting/debugging tips!

docs/docs/webui.md
mlunoe: So I experienced errors when I tried to compile the Chronos UI, specifically I got errors when in the minification step, some time back on my own machine (16 GB Memory). I think we can keep this in the documentation for now. And then we will fix these problems, when will start to rebuild the UI.

docs/resources.md
```diff
@@ -0,0 +1,17 @@
+---
+layout: narrow
+tab: resources
+title: Chronos Resources
+---
+
```
elingg: Nice links!

elingg:
Looks awesome, @ssk2! I would like to look it over together and if you could also address my comments, that would be great. After that, I can do a final review.

docs/docs/webui.md
elingg: Makes sense, we can add a disclaimer saying, "When building and optimizing the assets, make sure you have at least 1GB of RAM available" without linking to the issue.

docs/docs/api.md
ssk2: @elingg - what do you mean by the default configuration?

ssk2: Also, trying to steer clear of Mesosphere mentions since this repo is technically owned by the Mesos organisation.

elingg: Cool, we discussed this separately. It's using Chaos and ScallopConf which gives us by default 8080. About steering clear of Mesosphere mentions, that makes sense. If you check the license file though there is ownership by Airbnb and Mesosphere.

elingg:
:+1: LGTM, feel free to merge when ready as long as the build passes.

PR #447, opened 2015-05-13
Making paths relativ to enable proxied requests

aldipower:
This enables the Chronos UI to be run through proxied calls by making requests to the API and assets like images relative.
F.ex. http://dcos/chronos/#jobs/job1 is now working.

### DIFF
--- a/src/main/resources/assets/app/scripts/models/base_job.js
+++ b/src/main/resources/assets/app/scripts/models/base_job.js
@@ -21,7 +21,6 @@ function($, Backbone, _, moment, BaseJobValidations) {
         encoded;
 
     encoded = _.map(args, function(arg) { return encodeURIComponent(arg); });
-    encoded.unshift('');
     return encoded.join('/');
   }
 
--- a/src/main/resources/assets/app/scripts/views/job_detail_view.js
+++ b/src/main/resources/assets/app/scripts/views/job_detail_view.js
@@ -36,11 +36,9 @@ function($,
 
   'use strict';
 
-  var JobDetailView, asyncExecutorPath, Remove;
+  var JobDetailView, Remove;
 
   var LIST_SEPARATOR = ', ';
-
-  asyncExecutorPath = '/srv/mesos/utils/async-executor.arx';
 
   Remove = BoundView.prototype.remove;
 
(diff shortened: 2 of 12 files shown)


src/main/resources/assets/app/scripts/views/job_detail_view.js
```diff
@@ -36,12 +36,10 @@ function($,
 
   'use strict';
```
elingg: just a quick question @aldipower why was the asyncExecutorPath removed?

elingg:
LGTM! Any objections or comments, @philipnrmn or @brndnmtthws?

elingg:
Also, @mlunoe, please take a look

src/main/resources/assets/app/scripts/views/job_detail_view.js
aldipower: Very attentive. :)
asyncExecutorPath isn't used anywhere and is abandoned, so I decided to tidy up a little bit.

philipnrmn:
:shipit:

src/main/resources/assets/app/scripts/views/job_detail_view.js
elingg: Nice

src/main/resources/assets/app/scripts/models/base_job.js
```diff
@@ -21,7 +21,6 @@ function($, Backbone, _, moment, BaseJobValidations) {
         encoded;
 
     encoded = _.map(args, function(arg) { return encodeURIComponent(arg); });
-    encoded.unshift('');
```
mlunoe: I'm not sure what this did, or why it was removed?

mlunoe: It seems that is adding a fail-safe empty argument to the array, so if you have `[]`, it will work with `[""]`, which does not do any harm.

mlunoe:
Other than my comment, I think this looks good.

src/main/resources/assets/app/scripts/models/base_job.js
aldipower: This is not an empty fail-safe argument.
The only reason this was added is to let the .join('/') in the line below add the '/', so the array gets something like this ['', ''] what is joined to ''+'/'+''.
Now with relative paths, we don't need this anymore and the line was removed.

aldipower: Anyways, strange approach.. :)

mlunoe: Arh! I see now. Wow! That is weird!

PR #456, opened 2015-05-20
DOCKER: better entrypoint

anapsix:
As per Docker-Library best practices, it's best to  allow  overriding ENTRYPOINT, when needed by checking  for  "-"  arguments.
Examples:
`docker run -it chronos` is possible, and so is passing flags:
`docker run -it chronos --help` will work as expected / before.
`docker run -it chronos /bin/bash` and similar is now possible.

See https://github.com/docker-library/official-images/issues/692

### DIFF
--- a/bin/start-chronos.bash
+++ b/bin/start-chronos.bash
@@ -1,10 +1,9 @@
 #!/bin/bash
-set -o errexit -o pipefail
+set -o errexit -o nounset -o pipefail
 
 # allow overriding ENTRYPOINT
 # see https://github.com/docker-library/official-images/issues/692
-if [ -z "$1" ] || [ "${1:0:1}" == '-' ]; then
-  set -o errexit -o nounset -o pipefail
+if [ "${1:-absent}" == "absent" ] || [ "${1:0:1}" == '-' ]; then
   flags=( "$@" )
 
   heap=384m
--- a/bin/start-chronos.bash
+++ b/bin/start-chronos.bash
@@ -1,27 +1,39 @@
 #!/bin/bash
-set -o errexit -o nounset -o pipefail
+set -o errexit -o pipefail
 
-chronos_home="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd -P )"
-echo "Chronos home set to $chronos_home"
-export JAVA_LIBRARY_PATH="/usr/local/lib:/lib:/usr/lib"
-export LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-/lib}"
-export LD_LIBRARY_PATH="$JAVA_LIBRARY_PATH:$LD_LIBRARY_PATH"
+# allow overriding ENTRYPOINT
+# see https://github.com/docker-library/official-images/issues/692
+if [ -z "$1" ] || [ "${1:0:1}" == '-' ]; then
+  set -o errexit -o nounset -o pipefail
+  flags=( "$@" )
 
-flags=( "$@" )
+  heap=384m
+  chronos_home="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd -P )"
+  echo "Chronos home set to $chronos_home"
+  export JAVA_LIBRARY_PATH="/usr/local/lib:/lib:/usr/lib"
+  export LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-/lib}"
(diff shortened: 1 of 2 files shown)


bin/start-chronos.bash
```diff
@@ -1,27 +1,39 @@
 #!/bin/bash
-set -o errexit -o nounset -o pipefail
+set -o errexit -o pipefail
 
-chronos_home="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd -P )"
-echo "Chronos home set to $chronos_home"
-export JAVA_LIBRARY_PATH="/usr/local/lib:/lib:/usr/lib"
-export LD_LIBRARY_PATH="${LD_LIBRARY_PATH:-/lib}"
-export LD_LIBRARY_PATH="$JAVA_LIBRARY_PATH:$LD_LIBRARY_PATH"
```
elingg: To be cleaner here, this should be an if else

bin/start-chronos.bash
```diff
@@ -1,27 +1,39 @@
...
+# allow overriding ENTRYPOINT
+# see https://github.com/docker-library/official-images/issues/692
+if [ -z "$1" ] || [ "${1:0:1}" == '-' ]; then
+  set -o errexit -o nounset -o pipefail
+  flags=( "$@" )
 
-flags=( "$@" )
```
elingg: Why are we setting the heap here? If so, it needs to be configurable with a sensible default.

bin/start-chronos.bash
anapsix: +1 @elingg. To be cleaner here, second `set` is not required. I just wanted to make minimal changes to existing script.
But I had to remove `nounset` so that my if condition works when first argent is missing.

elingg: Hm, here's another idea. What about using ```set -o nounset``` inside the if statement?

anapsix: I don't see how it would make any difference @elingg, if you mean `if [ -z "$1" ]...` statement.
The _unset_, as I understand causes script to exit with error if any variable used is unset.. So what would it be catching there after if statement?

elingg: So this statement ```if [ -z "$1" ] || [ "${1:0:1}" == '-' ]``` will not work if args 0 and 1 don't exist. So ``` flags=( "$@" )``` should be fine if we get into the if statement without adding nounset. I think we should add ```set -o errexit -o nounset -o pipefail``` after the if statement you mentioned. The only thing it will be catching is ```exec "$@"``` at the end (assuming it didn't go into the if).

This is a nit, but I would like to be as careful as possible here.

anapsix: @eling `[ -z "$1"]` should work just fine when first arg is not present. That's why it is wrapped in double-quotes. `-z` is true when STRING length is zero, so when `$1` is not passed, bash would resolve it to empty string and as long as it's quoted, we are good. Unless of course, we have `nounset` set :). Arg 0 always exists.. though we don't really care.

elingg: What if we don't go into the if statement? We still want to use ```nounset``` for the final ```exec "$@"``` to make sure the at least 1 arg is defined, correct? Or am I missing something here?

anapsix: @elingg `$@` will success even is you pass no arguments to the script, though at that point `[ -z "$1" ]` will catch that, so the whole thing is moot.

Example:
```
#!/bin/bash
set -o nounset -o errexit -o pipefail
echo "Arguments: ${@}"     # will work if you pass no arguments with "-o nounset"
```

elingg: Thanks @anapsix for the long explanation. In that case, let's get rid of the second ```set``` as it seems to have no effect (again, please correct me if I'm missing something) and let's :ship: !

anapsix: well.. I suppose @guenter's intention for `set -o nounset` was to make sure script doesn't continue with any unset variables and as a general safeguard.
as often the case, there are many way of doing the same thing.. I'm leaving the original `set` as second line of the script and updating the first `if` statement to use string substitution, which is compatible with `-o nounset` and will not result in _unbound variable_ error
```
if [ "${1:-absent}" == "absent" ] || [ "${1:0:1}" == '-' ]; then
```

elingg: yep, I want to use that safeguard in all cases if possible (not just if we enter the if statement). Your fix looks good to me!

PR #462, opened 2015-06-03
Accept offers with role '*' with --mesos_role

felixb:
Chronos should accept offers for role '*' when running with a role other than '*'.


elingg:
LGTM, I'm not sure what changed, but I do see that Marathon also has code with ```r.getRole == "*"```. Let me know if there are any further objections.

modeyang:
LGTM, I also confronted that issue.

Issue #285, opened 2014-10-24
chronos REST api server doesn't response


chengweiv5:
We encountered this issue twice in about 3 months, generally we have about 25K jobs per day and the api server may doesn't response sometime. when this issue happen, `curl -X -L <host>:<port>/ping` will hang up for ever.

kolloch:
Hi @chengweiv5 , sorry for responding that late.

Did you ever do a stack trace when this happens? You can do that by issuing a `kill -3` or using the `jstack` command. It is helpful if you get multiple stack traces so that we can compare what changed or didn't change in-between.

Maybe, all threads accepting requests are blocked.

chengweiv5:
thanks for your tips, @kolloch, we'll do that if we lucky to reproduce this issue.

PR #463, opened 2015-06-04
Forbid using canBeReadonly when connecting zookeeper

depay:
Chronos may get a wrong leader path, which may cause a fatal error that:
once the non-leader instance gets an out-of-date leader path and the wrong
path happens to be itself, it will redirects a request to itself  and
then blocks other any requests in less than 1 second. In this circumstance,
scheduler is still running but no rest api responds, even it becomes leader.

So, setting connecting parameter 'canBeReadOnly' to false(default) to provide
Chronos from reading wrong leader info is necessary.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/config/ZookeeperModule.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/config/ZookeeperModule.scala
@@ -31,7 +31,7 @@ class ZookeeperModule(val config: SchedulerConfiguration with HttpConf)
   def provideZookeeperClient(): CuratorFramework = {
     val curator = CuratorFrameworkFactory.builder()
       .connectionTimeoutMs(config.zooKeeperTimeout().toInt)
-      .canBeReadOnly(true)
+      .canBeReadOnly(false)
       .connectString(validateZkServers())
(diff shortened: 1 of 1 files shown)


depay:
#285

elingg:
Thanks @depay! This change makes sense, and we need to fix this issue so I will merge. Question though, why does the leader path become correct? Shouldn't the HA be properly managed by Curator?

depay:
@elingg  it seems that something wrong with zookeeper at the very moment, chronos gets
the leader-changing event but cannot connect to the zookeeper leader, so it reads a leader path from another zk node. The strategy is supported by zookeeper but not curator.
~~~
canBeReadOnly - (added in 3.4) whether the created client is allowed to go
to read-only mode in case of partitioning. Read-only mode basically means
that if the client can't find any majority servers but there's partitioned server
it could reach, it connects to one in read-only mode, i.e. read requests are
allowed while write requests are not. It continues seeking for majority in the
background. 
~~~

modeyang:
LGTM,  thanks @depay,  the read-only flag need be false when leader switch.

elingg:
Thanks for the explanation, @depay. :ship:

PR #465, opened 2015-06-09
Remove persistence task after mesos task running

depay:
MesosDriver may crash during launching tasks, and it that case task is not
submitted to mesos but removed from persistence storage. Keep task util it
really executes on mesos.


elingg:
makes sense :+1:

PR #477, opened 2015-06-24
Add port_on_host for a visitable port on host

depay:
If Chronos run in Docker with --net=bridge, it should register visitable host:port on zookeeper for proxying requests.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/config/SchedulerConfiguration.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/config/SchedulerConfiguration.scala
@@ -113,7 +113,11 @@ trait SchedulerConfiguration extends ScallopConf {
   lazy val mesosAuthenticationSecretFile = opt[String]("mesos_authentication_secret_file",
     descr = "Mesos Authentication Secret",
     noshort = true)
-
+  // If Chronos run in Docker with --net=bridge, it should register visitable host:port on zookeeper for proxy.
+  lazy val portOnHost = opt[Int]("port_on_host",
+    descr = "Chronos port on host if using Docker BRIDGE network",
+    noshort = true,
+    default = None)
 
   def zooKeeperHostAddresses: Seq[InetSocketAddress] =
     for (s <- zookeeperServers().split(",")) yield {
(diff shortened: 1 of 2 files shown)


elingg:
I'd rather not create an extra variable if we can avoid it. Overriding the http-port doesn't work when running in bridge mode? If this is necessary, we definitely need a comment in the second file that was changed and updates to our documentation about this.

depay:
I mean that if we use some commands like `docker run -p 12345:8080 chronos-image`, other instance should proxy to "hostname:12345" instead of "hostname:8080", which should be register in zookeeper. "http_port" is used for exposing the port in container, and another variable for real port on host(e.g.12345) is required. Marathon will also deliver HOSTNAME and PORT(e.g.12345) to its container. It's a little ambiguous for the two variable. @elingg do you have any other solution?

elingg:
Thanks @depay, this is a tricky issue, because bridge networking for our docker images currently has a number of issues based on the network set up. If we do add this option, we do need to add the documentation for sure. I'm cc'ing @aquamatthias on here based on his knowledge of Marathon and how bridged networking works for Marathon docker containers.  He's also working on Chronos now as well!

felixb:
This is useful when putting chronos behind an reverse proxy too.

aquamatthias:
Hey @depay why can't you use the same port for chronos and the visible docker port? 
What is the reasoning to make this different?

depay:
since i wanna run chronos on marathon, which may cause a port conflict if "host" net used.

aquamatthias:
Hey @depay just start chronos with flag `--http_port 12345` and you are done.

depay:
how if another user on that slave also loves 12345 deeply...-_-#...

aquamatthias:
The PR suggests to define the advertised host port instead of the configured http port.
My proposal for this problem is: if you are running inside docker image in bridged mode, take the same container port as the external visible one. If you run your docker image in bridged mode, you should have full control over all the ports. The problem of port conflicts would arise on the host level and this problem is not addressed in your PR.

depay:
@aquamatthias it's just unreasonable to look for a suitable port to expose everytime building docker image from dockerfile. This PR just wanna supply a way to solve the probable question with diff ports. If the same port can always be suitable, of cause it's the best circumstance, user can only set the old param as current, and need not care about the new param, just leaving it default.
i know it's boring to add an extra variable but i've no other solution, under the current leader-follow proxy strategy.

depay:
@aquamatthias thanks & i've known how to run it through marathon now...just setting container port = 0 for a random port...

PR #486, opened 2015-07-06
Add LIKE constraint

wndhydrnt:
The PR adds a new constraint called LIKE which matches the value of an
attribute of a slave against a regular expression.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
@@ -1,0 +1,37 @@
+package org.apache.mesos.chronos.scheduler.jobs.constraints
+
+import java.util.logging.Logger
+
+import org.apache.mesos.Protos
+import scala.collection.JavaConversions._
+
+case class LikeConstraint(attribute: String, value: String) extends Constraint {
+
+  val regex = value.r
+
+  private[this] val log = Logger.getLogger(getClass.getName)
+
+  def matches(attributes: Seq[Protos.Attribute]): Boolean = {
+    attributes.find(a => a.getName == attribute).exists { a =>
+      a.getType match {
+        case Protos.Value.Type.SCALAR =>
+          regex.pattern.matcher(a.getScalar.getValue.toString).matches()
+        case Protos.Value.Type.SET =>
+          a.getSet.getItemList.exists(regex.pattern.matcher(_).matches())
+        case Protos.Value.Type.TEXT =>
+          regex.pattern.matcher(a.getText.getValue).matches()
+        case Protos.Value.Type.RANGES =>
+          log.warning("Like constraint does not support attributes of type RANGES")
+          false
+        case _ =>
+          val t = a.getType.getNumber
+          log.warning(s"Unknown constraint with number $t in Like constraint")
+          false
+      }
+    }
+  }
+}
+
+object LikeConstraint {
+  val OPERATOR = "LIKE"
+}
--- a/src/main/scala/org/apache/mesos/chronos/utils/JobDeserializer.scala
+++ b/src/main/scala/org/apache/mesos/chronos/utils/JobDeserializer.scala
@@ -1,5 +1,7 @@
 package org.apache.mesos.chronos.utils
 
+import java.util.regex.Pattern
+
 import org.apache.mesos.chronos.scheduler.config.SchedulerConfiguration
 import org.apache.mesos.chronos.scheduler.jobs._
 import org.apache.mesos.chronos.scheduler.jobs.constraints._
@@ -172,6 +174,8 @@ class JobDeserializer extends JsonDeserializer[BaseJob] {
         c.get(1).asText match {
           case EqualsConstraint.OPERATOR =>
             constraints.add(EqualsConstraint(c.get(0).asText, c.get(2).asText))
+          case LikeConstraint.OPERATOR =>
+            constraints.add(LikeConstraint(c.get(0).asText, c.get(2).asText))
           case _ =>
         }
       }
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -352,6 +352,18 @@ Schedule a job on nodes that share a common attribute.
 ...
 "constraints": [["rack", "EQUALS", "rack-1"]],
 ...
+```
 
+### LIKE constraint
+
+Schedule jobs on nodes which attributes match a regular expression.
+
+```json
+...
+"constraints": [["rack", "LIKE", "rack-[1-3]"]],
+...
+```
+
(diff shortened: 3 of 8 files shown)


wndhydrnt:
This patch is running in our cluster for about a month now and we never experienced any issues.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
```diff
@@ -0,0 +1,24 @@
+package org.apache.mesos.chronos.scheduler.jobs.constraints
+
+import org.apache.mesos.Protos
+
+case class LikeConstraint(attribute: String, value: String) extends Constraint {
+
+  val regex = value.r
+
+  def matches(attributes: Seq[Protos.Attribute]): Boolean = {
+    attributes.foreach { a =>
```
aquamatthias: Do not use foreach with return, but use find instead.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
```diff
@@ -0,0 +1,24 @@
...
+      if (a.getName == attribute) {
+        a.getText.getValue match {
+          case regex(_*) => return true
```
aquamatthias: Scala's internal regex is great for finding regexp groups etc.
For a simple matching filter, please use the simple form: `regex.pattern.matcher(value).matches()`

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
```diff
@@ -0,0 +1,24 @@
...
+import org.apache.mesos.Protos
+
+case class LikeConstraint(attribute: String, value: String) extends Constraint {
+
+  val regex = value.r
+
+  def matches(attributes: Seq[Protos.Attribute]): Boolean = {
+    attributes.foreach { a =>
+      if (a.getName == attribute) {
+        a.getText.getValue match {
```
aquamatthias: The attribute can contain multiple value types: scalar, ranges, set and text.

aquamatthias: The text field of an attribute is optional and can be null.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
```diff
@@ -0,0 +1,24 @@
+package org.apache.mesos.chronos.scheduler.jobs.constraints
+
+import org.apache.mesos.Protos
+
+case class LikeConstraint(attribute: String, value: String) extends Constraint {
+
+  val regex = value.r
```
aquamatthias: This will throw an exception for invalid regexp expression (e.g. `"[[[".r` )

src/main/scala/org/apache/mesos/chronos/utils/JobDeserializer.scala
```diff
@@ -172,6 +172,8 @@ class JobDeserializer extends JsonDeserializer[BaseJob] {
         c.get(1).asText match {
           case EqualsConstraint.OPERATOR =>
             constraints.add(EqualsConstraint(c.get(0).asText, c.get(2).asText))
+          case LikeConstraint.OPERATOR =>
+            constraints.add(LikeConstraint(c.get(0).asText, c.get(2).asText))
```
aquamatthias: This will fail, if the constraint array does not hold 3 elements.

aquamatthias:
@wndhydrnt Thanks for this PR.
Generally looks good. I added some comments. 
From the user perspective I would like to have some validation during creation time, that tells me about wrong constraint definitions, wrong regexp definition etc.
@elingg do we have validation logic in place? Or do we rely on exceptions thrown during creation time?

elingg:
hi @aquamatthias! I agree with your review here. To answer your questions, the validation logic in this project could be improved, as there is definitely reliance on exceptions thrown during creation time. See https://github.com/mesos/chronos/blob/ff3fbdfc2e55ace2dc0b8cd3d45adf8853eea33a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Iso8601Expressions.scala, and you will see a lack of validation logic for instance.  I think more of that as well as error handling is always better, and we should do our best to iteratively improve that in this project.

aquamatthias:
@elingg: Thanks for explanation. 
@wndhydrnt: As long as we there is no validation logic in place, the existing mechanism of creating the LikeConstraint (JobDeserializer and eager regex creation) should be sufficient.

wndhydrnt:
@aquamatthias I've updated the code according to your suggestions.

The LikeConstraint now supports scalar, text and set. I could not think of a valid use for ranges so I left it out for now. I am also not sure whether supporting set is a good idea. Sets, in general, could be supported by adding another constraint called, for example, `CONTAINS`.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def matches(attributes: Seq[Protos.Attribute]): Boolean = {
+    attributes.exists { a =>
+      if (a.getName == attribute) {
+        a.getType match {
+          case Protos.Value.Type.SCALAR =>
+            return regex.pattern.matcher(a.getScalar.getValue.toString).matches()
```
aquamatthias: Please don't use return - this is only a short circuit that is not needed here, since exist will return directly after an element is found.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/constraints/LikeConstraint.scala
```diff
@@ -0,0 +1,30 @@
...
+          case Protos.Value.Type.SET =>
+            return a.getSet.getItemList.exists(regex.pattern.matcher(_).matches())
+          case Protos.Value.Type.TEXT =>
+            return regex.pattern.matcher(a.getText.getValue).matches()
+          case _ =>
```
aquamatthias: log a warning here, since you don't expect any different type.

aquamatthias:
Thanks for this update.
I just had some small comments on this - but generally this is looking good!
It would be nice, if you could also add a test case for wrong specified LIKE constraints, 
so that we are sure that will fail. 
If you also could squash the commit, we then can merge that into master.

Great enhancement!

elingg:
:+1:

wndhydrnt:
@aquamatthias Improved the code, added logging and specs and also a note in the docs to make it clear which attributes are supported by the LIKE constraint.

aquamatthias:
@wndhydrnt Great enhancement. Thanks.

PR #489, opened 2015-07-07
Added Overview to doc table of contents

joel-hamill:
This will allow Overview to show in main doc view, along with getting started.


aquamatthias:
Thanks @joel-hamill

elingg:
Should we also merge this change into the gh-pages branch or was this already done, @aquamatthias / @joel-hamill ?

joel-hamill:
@elingg I think we should. Is there additional action I need to take?

elingg:
@joel-hamill, Yes, you need to check in those changes to the gh-pages branch. Glad to help or @ssk2 or @aquamatthias can show you.

PR #491, opened 2015-07-08
Clarify arguements param on manual job starts.

### LINKED ISSUE #490: Example in "Manually Starting a Job" doesn't work
The "Manually Starting a Job" section on http://mesos.github.io/chronos/docs/api.html recommends using the 'arguments' parameter, but doesn't mention that 'shell' must be set to false for it to work. This should be clarified/fixed.

KyleJamesWalker:
Fixes #490.

The "Manually Starting a Job" section on http://mesos.github.io/chronos/docs/api.html recommends using the 'arguments' parameter, but doesn't mention that 'shell' must be set to false for it to work. This should be clarified/fixed.


gkleiman:
Thanks! :+1:

PR #499, opened 2015-07-17
Made doc links use https

solarkennedy:
The chronos docs look strange on http because of a lack of js on one of the cdns. This fixes that.
Also https is good.


gkleiman:
Thanks a lot!

PR #503, opened 2015-07-29
Update Chronos to allow to start a job with custom arguments

doronin:
Current commit re-introduces the the ability to pass job arguments to a force run from https://github.com/mesos/chronos/pull/337 which was lost in https://github.com/mesos/chronos/pull/360/files commit.

PR360/355 fixed the issue with `null` arguments (https://github.com/mesos/chronos/issues/355). I verified that existing functionality (https://mesos.github.io/chronos/docs/api.html#manually-starting-a-job) works without introducing `null` arguments when optional `JobParams` payload is omitted from request

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
     <!-- runtime deps versions -->
     <akka.version>2.3.6</akka.version>
     <cassandra-driver-core.version>2.1.0</cassandra-driver-core.version>
-    <chaos.version>0.6.5</chaos.version>
+    <chaos.version>0.6.3</chaos.version>
     <commons-math3.version>3.2</commons-math3.version>
     <curator-framework.version>2.6.0</curator-framework.version>
     <guava.version>16.0.1</guava.version>
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
     <!-- runtime deps versions -->
     <akka.version>2.3.6</akka.version>
     <cassandra-driver-core.version>2.1.0</cassandra-driver-core.version>
-    <chaos.version>0.6.3</chaos.version>
+    <chaos.version>0.6.5</chaos.version>
     <commons-math3.version>3.2</commons-math3.version>
     <curator-framework.version>2.6.0</curator-framework.version>
     <guava.version>16.0.1</guava.version>
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
@@ -142,14 +142,17 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
   @Path(PathConstants.jobPatternPath)
   @PUT
   @Timed
-  def trigger(@PathParam("jobName") jobName: String,
-              @QueryParam("arguments") arguments: String
-               ): Response = {
+  def trigger(@PathParam("jobName") jobName: String, jobParams: JobParams): Response = {
     try {
       require(jobGraph.lookupVertex(jobName).isDefined, "Job '%s' not found".format(jobName))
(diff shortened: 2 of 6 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -142,14 +142,21 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
...
+  def trigger(@PathParam("jobName") jobName: String, jobParams: JobParams): Response = {
     try {
       require(jobGraph.lookupVertex(jobName).isDefined, "Job '%s' not found".format(jobName))
       val job = jobGraph.getJobForName(jobName).get
       log.info("Manually triggering job:" + jobName)
-      jobScheduler.taskManager.enqueue(TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0), job.highPriority)
+
+      var extraJobArguments: Seq[String] = List()
+
+      if (jobParams!=null) {
```
kolloch: Using `Option(jobParams).map(_.arguments).getOrElse(List.empty)` would be more idiomatic, in my opinion.

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -142,14 +142,21 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
...
+        extraJobArguments = jobParams.arguments
+        log.info("JobParams.arguments: " + extraJobArguments.toString())
+      }
+
+      val taskId = TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0, extraJobArguments)
```
kolloch: So the only effect here is that the taskId includes the parameters?

pom.xml
```diff
@@ -23,7 +23,7 @@
     <!-- runtime deps versions -->
     <akka.version>2.3.6</akka.version>
     <cassandra-driver-core.version>2.1.0</cassandra-driver-core.version>
-    <chaos.version>0.6.3</chaos.version>
+    <chaos.version>0.6.5</chaos.version>
```
kolloch: Shouldn't that be a different pull request?

doronin: you're right. that change comes from https://github.com/doronin/chronos/commit/e55eaa92e8743776f987ec7c70a4d06d23364114 and got here via `git pull --rebase`

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
doronin: correct

src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
doronin: Ok. I'll update the code

doronin:
Peter I removed changes to `pom.xml` and refactored JobManagementResource to use more idiomatic style of setting `extraJobArguments`

doronin:
@kolloch do you have any ETA of merging the code in this PR? We really need this feature restored. Thanks!

doronin:
@kolloch @elingg hey guys, do you know when I can expect to see this PR land in release? Thanks

kolloch:
Hi @doronin,

I think you did not really reintroduce the functionality because

* You changed the API (body instead of a query parameter)
* and as far as I can see the arguments only affect the taskID and not the arguments of the executed command

For us to accept the patch we would also need tests that exercise this functionality.

djamelz:
It would be good to close this PR since this merged PR make the job : https://github.com/mesos/chronos/pull/717

brndnmtthws:

brndnmtthws:

PR #507, opened 2015-08-04
Add list of command line arguments to documentation

tdooner:
Hey, thanks for all the great work on Chronos.

I find some parts of the documentation impossible to use. Especially as Chronos is such a cornerstone of DCOS, more and more users like myself will be searching for how to configure chronos.

This adds a quick dump of command line flags to the Configuring page. I know a similar list was removed in 9edee8457847e380e5bc822f6d27854616299e25, but I think it is unacceptable to instruct users to install (or worse, build) your software in order to see the list of configuration options.

Thanks again.


kolloch:
I agree. Thanks!

PR #510, opened 2015-08-06
Fixes #509 - Flag for reviveOffers and the duration for which to reject offers

### LINKED ISSUE #509: Chronos should set a non-default filter when declining offers
**Use case:** _As a cluster operator I would like to run many frameworks (possibly multiple Chronos instances) on the same Mesos cluster.  I'd like to mitigate offer starvation; that is, schedulers with work to do should receive offers from the Mesos master._

From the SchedulerDriver documentation, Chronos can use both the overloaded definition of `declineOffer` to set filters:

```java
  /**
   * Declines an offer in its entirety and applies the specified
   * filters on the resources (see mesos.proto for a description of
   * Filters). Note that this can be done at any time, it is not
   * necessary to do this within the {@link Scheduler#resourceOffers}
   * callback.
   *
   * @param offerId The ID of the offer to be declined.
   * @param filters The filters to set for any remaining resources.
   *
   * @return        The state of the driver after the call.
   *
   * @see OfferID
   * @see Filters
   * @see Status
   */
  Status declineOffer(OfferID offerId, Filters filters);
```

and clear them later on.

```
  /**
   * Removes all filters, previously set by the framework (via {@link
   * #launchTasks}). This enables the framework to receive offers
   * from those filtered slaves.
   *
   * @return    The state of the driver after the call.
   *
   * @see Status
   */
  Status reviveOffers();
```

**Proposal:** _When the task queue becomes empty, decline subsequent offers with a long timeout (say, max double value).  When the task queue becomes non-empty, invoke reviveOffers.

**NOTE:** This issue is analogous to https://github.com/mesosphere/marathon/issues/1931

gkleiman:
@ConnorDoyle @elingg @brndnmtthws 

**NOTE**: I plan to add some tests tomorrow, but I would really appreciate if I could get any feedback in the meantime. This has already been done to Marathon, please see https://github.com/mesosphere/marathon/issues/1931.

Add the following command line flags:

`--revive_offers_for_new_jobs` if true, revive offers is called when a new job
is added to the `TaskManager`.

Also note, that Mesos only filters offers that are a strict sub set of a
rejected offer.

`--decline_offer_duration` allows configuring the duration for which unused
offers are declined.

`--min_revive_offers_interval` if `--revive_offers_for_new_jobs` is specified,
do not call reviveOffers more often than this interval. It defaults to 5
seconds.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
@@ -8,9 +8,10 @@ import com.google.common.cache.CacheBuilder
 import com.google.common.util.concurrent.{ListenableFutureTask, ListeningScheduledExecutorService}
 import com.google.inject.Inject
 import org.apache.mesos.Protos.{TaskID, TaskState}
+import org.apache.mesos.chronos.scheduler.config.SchedulerConfiguration
 import org.apache.mesos.chronos.scheduler.graph.JobGraph
 import org.apache.mesos.chronos.scheduler.jobs.stats.{CurrentState, JobStats}
-import org.apache.mesos.chronos.scheduler.mesos.MesosDriverFactory
+import org.apache.mesos.chronos.scheduler.mesos.{ MesosDriverFactory, MesosOfferReviver }
 import org.apache.mesos.chronos.scheduler.state.PersistenceStore
 import org.joda.time.{DateTime, DateTimeZone}
 
@@ -26,7 +27,9 @@ class TaskManager @Inject()(val listeningExecutor: ListeningScheduledExecutorSer
                             val jobGraph: JobGraph,
                             val mesosDriver: MesosDriverFactory,
                             val jobsObserver: JobsObserver.Observer,
-                            val registry: MetricRegistry) {
+                            val registry: MetricRegistry,
+                            val config: SchedulerConfiguration,
+                            val mesosOfferReviver: MesosOfferReviver) {
 
   val log = Logger.getLogger(getClass.getName)
 
@@ -160,6 +163,10 @@ class TaskManager @Inject()(val listeningExecutor: ListeningScheduledExecutorSer
         val (_, _, attempt, _) = TaskUtils.parseTaskId(taskId)
         val job = jobOption.get
         jobsObserver.apply(JobQueued(job, taskId, attempt))
+    }
+
+    if (config.reviveOffersForNewJobs()) {
+      mesosOfferReviver.reviveOffers()
     }
   }
 
--- a/pom.xml
+++ b/pom.xml
@@ -36,6 +36,7 @@
     <raven.version>4.1.2</raven.version>
 
     <!-- test deps versions -->
+    <akka-testkit.version>2.3.6</akka-testkit.version>
     <junit.version>4.11</junit.version>
     <mockito.version>1.9.5</mockito.version>
     <specs2.version>2.3.10</specs2.version>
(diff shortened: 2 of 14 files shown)


ConnorDoyle:
One important thing to add is to `reviveOffers` on registration and re-registration.  See: https://github.com/mesosphere/marathon/compare/d3560b4e52330ac63b152b667eca1ea78c41bfac...marathon-0.9.1-backports#diff-27dd4ebda88590d649eae0d05e3a6b47R34

gkleiman:
@ConnorDoyle: thanks for reviewing this.

We do that here: https://github.com/mesos/chronos/pull/510/files#diff-3b3cdf7fd755c23f1c70baa2d7c90368R49

ConnorDoyle:
Oops, I missed it.  Great!

gkleiman:
Added some unit tests.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskManager.scala
```diff
@@ -161,6 +164,10 @@ class TaskManager @Inject()(val listeningExecutor: ListeningScheduledExecutorSer
         val job = jobOption.get
         jobsObserver.apply(JobQueued(job, taskId, attempt))
     }
+
+    if (config.reviveOffersForNewJobs()) {
+      mesosOfferReviver.reviveOffers()
```
brndnmtthws: Maybe add a counter here? From the metrics registry.

brndnmtthws:
Code looks good. Can you comment on how this has been tested? Aside from the unit tests (thanks for those, btw).

gkleiman:
@brndnmtthws 

Thanks for having taken the time to review this.

The new actor debounces reviveOffers calls, so I added two counters: `reviveOffersRequest` and `reviveOffers`. The first one counts the number of times a reviveOffers was requested and the latter the number of times reviveOffers was actually called on the Mesos driver.

I added a couple more unit tests and also did some manual testing:

### Reproducing the starvation
* Start Mesos locally: `MESOS_RESOURCES="cpus(*):1;mem(*):512;disk(*):100" mesos local --ip=127.0.0.1`
* Start 10 Chronos frameworks without the new flags:

```shell
#!/bin/sh

i=0;

while [[ $i -le 9 ]]; do
  bin/start-chronos.bash --master 127.0.0.1:5050 \
    --graphite_host_port 192.168.99.100:2003 \
    --graphite_reporting_interval 5 \
    --graphite_group_prefix "chronos_${i}" \
    --http_port "808${i}" \
    --zk_path "/chronos${i}/state" &>"/tmp/chronos_logs/${i}.log"&

  i=$((i+1));
done;
```
* Add the following jobs to one of the just-started Chronos frameworks:

```json
{
   "name": "test",
   "command": "sleep 500000",
   "schedule": "R//PT10M",
   "runAsUser": "gaston",
   "disk": "35",
   "mem": "1"
}
```

```json
{
   "name": "starved",
   "command": "sleep 20",
   "schedule": "R//PT1M",
   "runAsUser": "gaston",
   "disk": "2",
   "mem": "1"
}
```

* Watch in the logs how that framework stops getting offers after having scheduled `test` and is unable to start `starved`.

### Trying out the new flags
* Stop all the processes and clean up the ZK state.
* Start Mesos as before.
* Start the 10 Chronos frameworks, this time with the following extra parameters: `--revive_offers_for_new_jobs  --decline_offer_duration 3000000`.
* Add the same jobs as before.
* Verify that the framework is able to start both the `test` and the `starved` job.

brndnmtthws:
Looks great to me, nice work. :+1: 

Merge away.

PR #515, opened 2015-08-10
remove task from persistence store on completion

meelapshah:
Old tasks can take up a lot of space in the persistence store, and can cause chronos to take a long time to start up as it cleans them up. It seems nicer to remove these tasks from the persistence store when they complete.


brndnmtthws:
Makes sense.

Can you comment on how this has been tested?

meelapshah:
I've been using this patch for a few months in my own setup without issue, but I haven't written any tests.

brndnmtthws:
Excellent, thanks!

PR #518, opened 2015-08-12
Issue 473: Generic Http callback for failed jobs in Chronos

kapil-malik:


src/main/scala/org/apache/mesos/chronos/notification/HttpClient.scala
```diff
@@ -0,0 +1,93 @@
...
+class HttpClient(val endpointUrl: String, 
+                 val credentials: Option[String]) extends NotificationClient {
+
+  private[this] val log = Logger.getLogger(getClass.getName)
+
+  def sendNotification(job: BaseJob, to: String, subject: String, message: Option[String]) {
+
+    val jsonBuffer = new StringWriter
+    val factory = new JsonFactory()
+    val generator = factory.createGenerator(jsonBuffer)
```
brndnmtthws: We use Jackson bindings elsewhere in the code. Can you do that here too please?

brndnmtthws:
Seems like a good idea.

Can you please describe how it's been tested? The addition of unit tests would be appreciated as well.

src/main/scala/org/apache/mesos/chronos/notification/HttpClient.scala
kapil-malik: Hi @brndnmtthws 
I borrowed the JSON generation code from https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/notification/SlackClient.scala.
Do you have plans to change that as well?

kapil-malik:
Hi @brndnmtthws 
I've manually tested this by setting up a small REST endpoint listening to a generic POST request. The endpoint received failure notifications from chronos with all the details.
I could not find any unit-tests for existing notification clients defined in https://github.com/mesos/chronos/tree/master/src/main/scala/org/apache/mesos/chronos/notification like SlackClient, RavenClient etc.
Would you suggest some approach on how to add unit-tests for notification clients?

kapil-malik:
Hi @brndnmtthws  / all,

Can one of you kindly accept this pull request, considering the use case is genuine (used by my team right now) and generic. Further, the code (libraries etc.) is in line with other notification clients.

Thanks and regards,

brndnmtthws:
Thanks for the patch.

kapil-malik:
Thanks for merging, this feature is quite useful to us.

PR #521, opened 2015-08-14
Fixes #520 - fetch the frameworkId from the state store on MesosDriver creation

### LINKED ISSUE #520: Chronos sometimes registers using an empty frameworkId after a leader fail over
After the network partition it has been observed (#511) that the initial Chronos leader fails over as expected, but the new Chronos leader does a registration with an empty frameworkId, instead of using the frameworkId obtained by the initial leader.

This results in a registration instead of a re-registration.

gkleiman:


src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosDriverFactory.scala
```diff
@@ -1,59 +1,47 @@
...
 
 /**
  * The chronos driver doesn't allow calling the start() method after stop() has been called, thus we need a factory to
  * create a new driver once we call stop() - which will be called if the leader abdicates or is no longer a leader.
  * @author Florian Leibert (flo@leibert.de)
  */
-class MesosDriverFactory(val mesosScheduler: Scheduler, val frameworkInfo: FrameworkInfo, val config: SchedulerConfiguration) {
+class MesosDriverFactory(
+                          val scheduler: Scheduler,
+                          val frameworkIdUtil: FrameworkIdUtil,
```
kolloch: why is this a `val`? We should avoid exporting dependencies as fields whenever possible.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosDriverFactory.scala
```diff
@@ -1,59 +1,47 @@
...
+                          val config: SchedulerConfiguration with HttpConf,
+                          val schedulerDriverBuilder: SchedulerDriverBuilder = new SchedulerDriverBuilder) {
 
   private[this] val log = Logger.getLogger(getClass.getName)
 
   var mesosDriver: Option[SchedulerDriver] = None
 
-  def start() {
+  def start() = {
```
kolloch: If you use this syntax, please specify the return type.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosDriverFactory.scala
```diff
@@ -1,59 +1,47 @@
...
-        new MesosSchedulerDriver(mesosScheduler, frameworkInfo, config.master(), credential)
-      case None =>
-        new MesosSchedulerDriver(mesosScheduler, frameworkInfo, config.master())
-    }
-
-    mesosDriver = Option(driver)
-  }
-
-  def close() {
+  def close() = {
```
kolloch: See above. `: Unit`

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosDriverFactory.scala
```diff
@@ -1,59 +1,47 @@
...
     assert(mesosDriver.nonEmpty, "Attempted to close a non initialized driver")
     if (mesosDriver.isEmpty) {
+      log.severe("Attempted to close a non initialized driver")
```
kolloch: Thanks for this ;)

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosDriverFactory.scala
```diff
@@ -1,59 +1,47 @@
...
       System.exit(1)
```
kolloch: Maybe use different exit codes for different error conditions?

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/SchedulerDriverBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+    else {
+      // ssl disabled, use http
+      frameworkInfoBuilder.setWebuiUrl(s"http://${config.hostname()}:${config.httpPort()}")
+    }
+
+    // set the authentication principal, if provided
+    config.mesosAuthenticationPrincipal.get.foreach(frameworkInfoBuilder.setPrincipal)
+
+    val frameworkInfo = frameworkInfoBuilder.build()
```
kolloch: Maybe extract building the frameworkInfo into its own local method?

src/test/scala/org/apache/mesos/chronos/ChronosSpec.scala
```diff
@@ -0,0 +1,18 @@
+package org.apache.mesos.chronos
+
+import mesosphere.chaos.http.HttpConf
+import org.apache.mesos.chronos.scheduler.config.SchedulerConfiguration
+import org.rogach.scallop.ScallopConf
+import org.specs2.mock.Mockito
+import org.specs2.mutable.SpecificationWithJUnit
+
+trait ChronosSpec extends SpecificationWithJUnit with Mockito {
+  def makeConfig(args: String*): SchedulerConfiguration with HttpConf = {
```
kolloch: I do not particularly like inheritance for getting additional methods in scope. This method could simply be part of a ChronosTestHelper object.

src/test/scala/org/apache/mesos/chronos/scheduler/mesos/MesosDriverFactorySpec.scala
```diff
@@ -0,0 +1,26 @@
...
+        frameworkIdUtil,
+        makeConfig(),
+        mock[SchedulerDriverBuilder])
+
+      frameworkIdUtil.fetch(any, any).returns(None)
+
+      mesosDriverFactory.get()
+
+      there was one(frameworkIdUtil).fetch(any, any)
+      ok
```
kolloch: shouldn't we test that it is also stored when we generated a new one?

kolloch: Ah, it's stored in the registered callback. Makes sense.

gkleiman:
@kolloch: thanks for your review comments! I addressed them and updated the PR.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/SchedulerDriverBuilder.scala
```diff
@@ -0,0 +1,89 @@
...
+      }
+      catch {
+        case cause: Throwable =>
+          throw new IOException(s"Error reading authentication secret from file [$secretFile]", cause)
+      }
+
+      credentialBuilder.build()
+    }
+
+    def buildFrameworkInfo(
```
kolloch: wrapping strange

kolloch:
LGTM

PR #525, opened 2015-08-18
Set the hostname in the framework information proto

gkleiman:
Due to a bug in Mesos < 0.24.0, Marathon users have observed JVM crashes
when the framework the host's hostname is nor resolvable
(https://issues.apache.org/jira/browse/MESOS-3145).

This change works around the problem by setting the hostname in the
Framework Information proto.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/config/SchedulerConfiguration.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/config/SchedulerConfiguration.scala
@@ -37,8 +37,8 @@ trait SchedulerConfiguration extends ScallopConf {
     descr = "The list of ZooKeeper servers for storing state",
     default = Some("localhost:2181"))
   lazy val hostname = opt[String]("hostname",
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/config/SchedulerConfiguration.scala
```diff
@@ -37,8 +37,8 @@ trait SchedulerConfiguration extends ScallopConf {
     descr = "The list of ZooKeeper servers for storing state",
     default = Some("localhost:2181"))
   lazy val hostname = opt[String]("hostname",
-    descr = "The advertised hostname stored in ZooKeeper so another standby " +
-      "host can redirect to this elected leader",
+    descr = "The advertised hostname, stored in ZooKeeper so another standby " +
```
kolloch: My proposal:

"The advertised hostname of this Chronos instance for network communication. This is used by other Chronos instances and the Mesos master."

kolloch:
I feel blown away... :+1:

PR #535, opened 2015-08-25
Allow periods in job names

carroux:
job_name_fail

Marathon [allows](https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md#id-string) periods in job names, but Chronos rejects jobs with periods in their names. Periods are useful as a delimiter, and there's no obvious reason why they aren't allowed in job names. I updated the regex to allow periods and added a test to check that the job name validation accepts a job name containing periods.

Tests [ran](https://travis-ci.org/carroux/chronos/builds/77242330#L3941) and [passed](https://travis-ci.org/carroux/chronos/builds/77242330#L3946).

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobUtils.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobUtils.scala
@@ -19,7 +19,7 @@ import scala.collection.mutable.ListBuffer
  */
 object JobUtils {
 
-  val jobNamePattern = """([\w\s#_-]+)""".r
+  val jobNamePattern = """([\w\s\.#_-]+)""".r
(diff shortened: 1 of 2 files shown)


keshavdv:
I'm looking forward to this!

felixb:
How about /? 
I'd love to have the same hierarchy as in marathon.

mrtyler:
shipit

FWIW @felixb I am in favor of allowing / and a bunch of other standard ascii separator characters but I have no idea if there are constraints the Chronos authors were trying to follow when they decided this regex (and I don't see any comments about it in JobUtils.scala). Maybe one of them will chime in :)

brndnmtthws:
Looks great, thanks!

PR #539, opened 2015-08-26
Fix Dockerfile so node works with Ubuntu base image

potto007:
Removed node package as it is a radio package in Debian/Ubuntu. Added symlink /usr/bin/node pointing at /usr/bin/nodejs.


brndnmtthws:
Thanks!

potto007:
No problem, Brenden!

PR #545, opened 2015-09-03
Document the job name requirements

### LINKED ISSUE #495: Expose chronos job name regex in docs
https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobUtils.scala#L22 defines the regex that an application name must match.  Please document this explicitly in the user docs.  Note that marathon already does this https://github.com/mesosphere/marathon/blob/master/src/main/resources/mesosphere/marathon/api/v2/AppsResource_create.md#id

carroux:
Closes #495

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -84,7 +84,7 @@ You can manually start a job by issuing an HTTP request.
 The heart of job scheduling is a JSON POST request.
 The JSON hash you send to Chronos should contain the following fields:
 
-* `name`: The job name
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Can you update this to reflect the change from #535?

Thanks!

carroux:
@brndnmtthws: Done!

brndnmtthws:
Beautiful, thanks!

PR #551, opened 2015-09-15
Fixes #540 - Document Search Endpoint

### LINKED ISSUE #540: Extend documentation of /scheduler/jobs by the possibility to get a single job
How do I get the definition of a single job: 
``` /scheduler/jobs/search?name=<job_name> ```

### PR DESCRIPTION
**Pull request by mwasn**

### CODE REVIEW

File: docs/docs/api.md [hunk 1]
```diff
@@ -49,6 +50,16 @@ Interesting fields in the hashes are:
 
 If there is a `parents` field there will be no `schedule` field, and vice-versa.
 
+## Searching for a Job
+
+Searching for a job with known attributes can be helpful when a large amount of jobs are being deployed. The list of possible attributes is described in the [Job Configuration](#job-configuration) section.
+
```
gkleiman: Not all of those attributes are supported by the search, see: https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala#L221

The search endpoints accepts the following parameters:

- name
- command
- any (which is quite misleading as it actually means "name or command").

We should also mention how the search works. It will convert both the search term and the desired job attribute to lower case and then check if the job attribute contains the term.

docs/docs/api.md
```diff
@@ -49,6 +50,16 @@ Interesting fields in the hashes are:
...
+* Endpoint: __/scheduler/jobs/search__
+* Method: __GET__
+* Example: `curl -L -X GET chronos-node:8080/scheduler/jobs/search?name=request_event_counter_hourly`
+* Example: `curl -L -X GET chronos-node:8080/scheduler/jobs/search?owner=foo@bar.io`
```
gkleiman: owner is not a supported parameter

mwasn: Changed

docs/docs/api.md
mwasn: Listed the available attributes and description.

kolloch:
Thanks! :fireworks:

PR #556, opened 2015-09-17
[docs:api] add clarity around values available for constraints

tony-kerz:


AtiX:
+1 thanks for pointing that out, would have saved me days if it was documented.

gkleiman:
Thanks for the pull request, and welcome!

PR #566, opened 2015-09-29
[pom] bump specs2

tony-kerz:


brndnmtthws:
Seems reasonable. Thanks.

Issue #570, opened 2015-10-02
docker build fails tests


tony-kerz:
currently running `docker build` against `master` will fail due to failing maven tests

```
SerializerAndDeserializer should::serialize and deserialize a ScheduleBasedJob correctly(org.apache.mesos.chronos.scheduler.api.SerDeTest)  Time elapsed: 0.115 sec  <<< FAILURE!
org.specs2.reporter.JUnitReporter$$anon$1: 'ScheduleBasedJob(FOO/BAR/BAM,A,noop,PT5M,10,20,fooexec,fooflags,7,foo@bar.com,Foo,Test schedule-based job,TODAY,YESTERDAY,true,0.1,256.0,128.0,false,0,ListBuffer(),false,,DockerContainer(dockerImage,ListBuffer(Volume(Some(/host/dir),container/dir,Some(RW)), Volume(None,container/dir,None)),ListBuffer(),HOST,true),,ListBuffer(EnvironmentVariable(FOO,BAR), EnvironmentVariable(AAAA,BBBB)),true,ListBuffer(-testOne),true,false,ListBuffer(EqualsConstraint(rack,rack-1), LikeConstraint(rack,rack-[1-3])))'
 is not equal to 
[...]

PR #571, opened 2015-10-02
tweaks to dockerfile and run script to facilitate alternate launch strategy

tony-kerz:
in my recent efforts working on a pull-request, i found it convenient to use `docker build` against the local `Dockerfile` to generate an image which i could subsequently reference in a marathon job along with the `bin/chronos-marathon` script to launch/evaluate a "SNAPSHOT" version of chronos in a mesos cluster.

there were a few things i needed to change to accomplish this:

(1) skip the tests in the maven directive per (https://github.com/mesos/chronos/issues/570)
(2) override a couple of default variables
(3) pass some additional command line parameters

this p/r encompasses those changes.

i didn't have the where-with-all to fix the existing test failures when running `docker build`, they are likely just environmental because it looks like the tests are succeeding in travis per the github hooks.

so i'm thinking it might be acceptable to skip tests when running `docker build`. 

if this is not the case, i can remove that commit if someone can get the tests to run within `docker build`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ ADD . /chronos
 
 WORKDIR /chronos
 
-RUN mvn clean package
+RUN mvn clean package -DskipTests
 
 EXPOSE 8080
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -16,7 +16,7 @@ ADD . /chronos
 
 WORKDIR /chronos
 
-RUN mvn clean package
```
gkleiman: My "docker build" run was successful. It seems the tests are flaky, but I think they should  be fixed rather than skipped.

So I'd remove this change from the PR.

brndnmtthws:
Looks like this is no longer applicable. Feel free to reopen if you want to revisit it.

PR #572, opened 2015-10-05
fix a formatting issue with api docs

rji:
Prior to this commit, there was a Markdown formatting problem with the
API docs. This resulted in a bunch of (what should have been) bullet
points appearing as normal text, and making the explanation of the ISO
8601 format harder to read than it needed to be.

This commit modifies some of the API documentation to correctly format
the bullet list for the "Adding a Scheduled Job" section.


rji:
Also note that this wasn't an issue with GitHub-flavored Markdown (i.e. viewing the docs in the repo), but that the API docs weren't formatted correctly on the Jekyll-generated site used for GitHub pages. It should now be appearing correctly both as GFM and the GitHub Pages site (tested locally using `jekyll serve`)

gkleiman:
Thank you for the PR!

rji:
Sure thing, @gkleiman! It looks like master needs to be synced with the gh-pages branch too.

PR #578, opened 2015-10-20
Fix the job listing lastSuccess and error type

orlandohohmeier:
Fix the job listing lastSuccess and lastError type. It's actually a date not a timestamp.


gkleiman:
Thanks!

PR #583, opened 2015-11-04
Add SCoverage for generating test coverage reports

spacejam:
@aquamatthias 
Let's get this into teamcity build reports.

### DIFF
--- a/docs/docs/contributing.md
+++ b/docs/docs/contributing.md
@@ -17,6 +17,9 @@ title: Contributor Guidelines
 - If the change is a bugfix, then the added tests must fail without the patch
   as a safeguard against future regressions.
 
+- Changes should not result in a drop in code coverage.  Coverage may be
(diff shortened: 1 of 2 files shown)


spacejam:
currently at 39.70%

kolloch:
I tested it. Thanks, it works. It would be great to document how to use it, though. I personally did not use maven much in the last ten years and I had to dig a bit until I came up with `mvn scoverage:check`.

kolloch:
@spacejam what do you think about adding short docs for that?

spacejam:
@kolloch I've added a note on coverage to the collaboration doc.  Do you think this is the right place for this?  I'm acting on my ulterior motive of encouraging test coverage across our projects to rise.

kolloch:
I don't know a better place to put it.

Thx! :ship: :rocket:

PR #595, opened 2015-12-01
add markJobSuccessful api call

Califax:
Add logic in job scheduler to trigger dependencies and Pull out logic for getting new job config with incremented success into method. Add unit tests

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -96,6 +96,19 @@ You can manually start a job by issuing an HTTP request.
 * Example: `curl -L -X PUT chronos-node:8080/scheduler/job/job_name?arguments=-debug`
 * Response: HTTP 204
 
+## Marking a job as successful
+
+You can manually mark a job as successful by issuing an HTTP request. If a job is marked successful, the success count
+of the job is incremented, the latest successful run time is updated, and all downstream dependencies are handled as if
+the job had completed executing the code in a standard run.
+the job normally runs.
+
+* Endpoint: ___/scheduler/job/success/<jobname>
+* Method: __PUT__
+* Query string parameters: `arguments` - jobname to be marked success
+* Example: `curl -L -X PUT chronos-node:8080/scheduler/job/success/request_event_counter_hourly`
+* Response: boolean (true or false depending on success of request)
+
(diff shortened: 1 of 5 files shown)


Califax:
Any chance I could get someone to take a quick look over this?

brndnmtthws:
Looks good, thanks!

docs/docs/api.md
```diff
@@ -96,6 +96,19 @@ You can manually start a job by issuing an HTTP request.
...
+You can manually mark a job as successful by issuing an HTTP request. If a job is marked successful, the success count
+of the job is incremented, the latest successful run time is updated, and all downstream dependencies are handled as if
+the job had completed executing the code in a standard run.
+the job normally runs.
+
+* Endpoint: ___/scheduler/job/success/<jobname>
+* Method: __PUT__
+* Query string parameters: `arguments` - jobname to be marked success
+* Example: `curl -L -X PUT chronos-node:8080/scheduler/job/success/request_event_counter_hourly`
+* Response: boolean (true or false depending on success of request)
```
ddossot: The response is actually a string formatted with: `"marked job %s as successful: %b"`

It is always `200 OK` even if `success` is false, which basically forces clients of the API to parse the returned string. Returning JSON would simplify parsing, using a different status code would be even better.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala
```diff
@@ -310,6 +299,43 @@ class JobScheduler @Inject()(val scheduleHorizon: Period,
...
+   * Mark job by job name as successful. Trigger any dependent children jobs that should be run as a result
+   */
+  def markJobSuccessAndFireOffDependencies(jobName : String): Boolean = {
+    val optionalJob = jobGraph.getJobForName(jobName)
+    if (optionalJob.isEmpty) {
+      log.warning("%s not found in job graph, not marking success".format(jobName))
+      return false
+    } else {
+      val job = optionalJob.get
+      jobMetrics.updateJobStatus(jobName, success = true)
```
ddossot: Shouldn't jobsObserver be notified here?

PR #603, opened 2015-12-09
Add uris to the CommandInfo when a custom executor is set

drexin:


gkleiman:
@drexin: thank you for the contribution!

PR #605, opened 2015-12-11
Don't allow invalid slash characters in ids

### LINKED ISSUE #568: Chronos should not add invalid characters to the Task ID
The `TaskUtils.getTaskId` method [adds the values of the arguments array](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala#L103) to the end of the task ID.

This makes for more-readable IDs in the case of common arguments like `-v`, but Mesos [considers `/` to be an invalid character in Task ID](https://github.com/apache/mesos/blob/a2e740d36550b1bf7f6fe9ad3ec0242a7f9d7a74/src/master/validation.cpp#L51). This prevents me from a job configuration like:

```json
"command": "cat",
"arguments": ["/etc/passwd"],
```

Mesos master (0.23.0) logs the following:

```
I1001 16:35:05.784687  5022 master.cpp:3792] Sending status update TASK_ERROR (UUID: [snip]) for task ct:1443717301964:0:my_chronos_job_name:cat /etc/password of framework 20150714-191408-4163031306-5050-1590-0001 'TaskID 'ct:1443717301964:0:my_chronos_job:cat /etc/passwd' contains invalid characters'
```

In my opinion, Chronos should either:

1. Strip invalid characters from the Task ID, or
2. Not add the arguments array to the Task ID in the first place

Since I don't see the benefit from adding the arguments, and the arguments can be quite long, I would prefer the second option. But whatever, I'd be happy with simply being able to use slashes in the arguments array. Thanks.

asottile:
Resolves #568

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
@@ -100,7 +100,7 @@ object TaskUtils {
   }
 
   def getTaskId(job: BaseJob, due: DateTime, attempt: Int = 0): String = {
-    taskIdTemplate.format(due.getMillis, attempt, job.name, job.arguments.mkString(" "))
+    taskIdTemplate.format(due.getMillis, attempt, job.name, job.arguments.mkString(" ").replaceAll("[/\\\\]", "_"))
   }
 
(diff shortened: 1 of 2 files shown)


asottile:
This almost certainly breaks something because of `TaskUtils.getJobArgumentsForTaskId`

Issue #597, opened 2015-12-01
Created jobs do not show in Internet Explorer


philipnrmn:
- Create a job in internet explorer
- It is not shown in the jobs list

Environment: IE: 11.0.9600.17905, on Windows Server 2012 R2 Datacenter

philipnrmn:
This is a CSS issue related to the window width:

[zoomy](https://s3.amazonaws.com/f.cl.ly/items/2Q0R3o3J063R0S371Q1m/Screen%20Recording%202015-12-17%20at%2010.49%20AM.gif?v=fec871f6)

PR #606, opened 2015-12-17
Prevent jobs list from collapsing in small screens

pierlo-upitup:
This fixes https://github.com/mesos/chronos/issues/597

fix

cc @mwasn


philipnrmn:
Sorry for the delay, and thanks @gkleiman for the reminder.

PR #615, opened 2016-01-11
Add port mappings

bengunter:
I based my changes off of 2.4.0 because that's the version we're using, and we needed this feature immediately.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
   <groupId>org.apache.mesos</groupId>
   <artifactId>chronos</artifactId>
-  <version>2.4.0-RC1</version>
+  <version>2.4.0</version>
   <inceptionYear>2012</inceptionYear>
 
   <prerequisites>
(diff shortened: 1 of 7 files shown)


brndnmtthws:
Can you rebase against master please?

andrebeat:
I think it would be cool if this also supported setting randomly assigned ports like in marathon (and especially keeping the same random port inside and outside the container) [[1]](https://mesosphere.github.io/marathon/docs/native-docker.html#bridged-networking-mode). Additionally, in order for this to be useful the assigned port should be exposed to the container (again like in marathon [[2]](https://mesosphere.github.io/marathon/docs/task-environment-vars.html#host-ports)).

brndnmtthws:
Could you rebase against master?

brndnmtthws:

bengunter:
@brndnmtthws We've moved to Singularity so I won't be following up on this. The code is here if anyone else wants to use it.

PR #621, opened 2016-01-21
Log a hint that constraints are not met for a task

solarkennedy:
I spent hours trying to debug a chronos job not working, but I was misled by this line because it said things like this:

```
WARN Insufficient resources remaining for task 'ct:1453417002396:0:paasta_canary sensu_canary git6c3833b5 config7708a392:', will append to queue. (Needed: [cpus: 0.1 mem: 256.0 disk: 256.0], Found: [cpus: 3.4499999990000014 mem: 22635.0 disk: 10000.0,cpus: 7.369999999999992 mem: 30363.0 disk: 10000.0,cpus: 2.5700000000000003 mem: 18051.0 disk: 10000.0,cpus: 7.349999999000012 mem: 3753.0...
```

And I was like, chronos man, looks like there are plenty of resources to me?

But it was a constraint issue.

This makes the error message hint at this, to benefit my future self?


gkleiman:
Thanks!

PR #626, opened 2016-02-01
Allow Chronos jobs to override taskInfo data

jamesmulcahy:
- Add 'taskInfoData' field to Chronos jobs
- If specified on a job, this data is placed in the
  data field of the taskInfo object, which is passed
  to the mesos executor.
- If not specified, then default Chronos task data
  is used (as per previous behaviour).


elingg:
LGTM

elingg:
James and I discussed and went through reviews and this is looking good. @jamesmulcahy, can you please rebase? 

Any objections or comments? If there are no objections, I will go ahead and merge.

Issue #627, opened 2016-02-02
Updating a job from scheduled to dependent leaves the job continuing to fire on original schedule


tyrannasaurusbanks:
Hi Guys,

I'm seeing some behaviour which I don't think is intended/appropriate, I thought I'd raise a issue here to see if it merits further investigation (or if I've got the wrong end of the stick).

It appears that updating a job from scheduled to dependent causes the original schedule to remain intact, resulting in the job firing twice: triggered once by the schedule, and once by the parent job completing successfully.

**Scenario**
I add the following 2 scheduled jobs:
```
$ curl -v -L -H "Content-Type: application/json" -X POST -d @testscheduledA.json http://localhost:8080/scheduler/iso8601
$ curl -v -L -H "Content-Type: application/json" -X POST -d @testscheduledB.json http://localhost:8080/scheduler/iso8601
```
```
{
  "schedule": "R/2016-01-20T15:40:00Z/PT24H",
  "name": "test_A",
  "epsilon": "PT30M",
  "command": "echo 'FOO' >> /tmp/JOB1_OUT",
  "owner": "bob@airbnb.com",
  "async": false
}
```
```
{
  "schedule": "R/2016-01-20T15:41:00Z/PT24H",
  "name": "test_B",
  "epsilon": "PT30M",
  "command": "echo 'FOO' >> /tmp/JOB1_OUT",
  "owner": "bob@airbnb.com",
  "async": false
}
```
I then update test_B to be a dependent job of test_A:
```
$ curl -v -L -H "Content-Type: application/json" -X POST -d @testdependentB.json http://localhost:8080/scheduler/dependency
```
testdependentB.json 
```
{
  "name": "test_B",
  "epsilon": "PT30M",
  "command": "echo 'FOO' >> /tmp/JOB1_OUT",
  "owner": "bob@airbnb.com",
  "async": false,
  "parents": [
    "test_A"
  ]
}
```

**Result**
The test_A job executes at the correct time, and kicks off test_B after it succeeds. But test_B then executes exactly 1 minute after test_A, which matches the original schedule set.

<img width="708" alt="screen shot 2016-02-02 at 3 43 34 pm" src="https://cloud.githubusercontent.com/assets/2912093/12754623/cf1dbcfa-c9c3-11e5-8904-31db8d8650a5.png">

[...]

PR #635, opened 2016-02-16
remove schedules from jobs made dependent when previously scheduled

Califax:


Califax:
https://github.com/mesos/chronos/issues/627

tyrannasaurusbanks:
:+1:

brndnmtthws:
Makes sense. Thanks!

PR #646, opened 2016-03-04
Chronos Update: Merge SIMS code with latest Chronos

wliu1labs:

PR #652, opened 2016-03-16
Chronos Update: Merge SIMS code with latest Chronos

wliu1labs:
With bug fix. Now Chronos is running on prod cluster in the correct way.


brndnmtthws:
Can you provide some background on what this PR is? How does it differ from the other 2 similarly named (#647 and #646) PRs?

PR #668, opened 2016-04-24
keep properties when a dependency job change to a schedule based job

jinzhao1994:
If job A is a schedule based job, job B is a dependency job. When A is removed from job list, job B will change to a schedule based job automatically. The code about this is use the construct function of schedule based job, pass each property of B. But it forget some property. Some of them are not very important, like description. Some of them are important, like constraint. I added all the property of base job.

Change-Id: Ide1d07eea04213969849a962bd8a96bc5be06897


jinzhao1994:
I have tested it in my computer. It runs well. My change is small and it has no relation with zookeeper. I don't know why it failed the test. I think it is right :P

brndnmtthws:

PR #671, opened 2016-04-26
Optimized Dockerfile

mikkokupsu:
Optimized Dockerfile to reduce image size by removing unnecessary dependencies and files


xiaods:
@mikkokupsu please rebase it.

PR #673, opened 2016-05-03
SerDeTest.scala: update tests to match values for mem, cpu and disk

pdericson:
mem, cpu and disk is set to 128.0, 0.1 and 256.0 here:

```
src/main/scala/org/apache/mesos/chronos/scheduler/config/SchedulerConfiguration.scala
```

mem, cpu and disk is set to 0.0, 0.0 and 0.0 here:

```
src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala
```

I suspect the non-zero defaults are only being set if ScheduleBasedJob is parsed from a strnig but I can't explain why we're seeing these failures now or for how long they have been present.


brndnmtthws:

PR #674, opened 2016-05-03
Optimizing Dockerfile

mikkokupsu:
Reducing the image size by removing unnecessary dependencies.


brndnmtthws:
It looks like this is no longer relevant.

PR #703, opened 2016-07-11
Change default framework name

mattmb:
This removes the version from the default framework name. This brings Chronos inline with other frameworks that don't include the version number in this field.

I hope there's no reason why Chronos currently includes the version number, please let me know if that's the case!


brndnmtthws:
Please rebase against master.

brndnmtthws:
Makes sense. Could you rebase against master?

mattmb:
Sorry must have missed the notification last time! Just rebased and pushed 😄

PR #717, opened 2016-08-23
task arguments

kensipe:
enabling the ability to have the REST API override or add arguments to a job.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
@@ -151,7 +151,8 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
       require(jobGraph.lookupVertex(jobName).isDefined, "Job '%s' not found".format(jobName))
       val job = jobGraph.getJobForName(jobName).get
       log.info("Manually triggering job:" + jobName)
-      jobScheduler.taskManager.enqueue(TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0), job.highPriority)
+      jobScheduler.taskManager.enqueue(TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0, Option(arguments).filter(_.trim.nonEmpty))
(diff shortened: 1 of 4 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/api/JobManagementResource.scala
```diff
@@ -151,7 +151,8 @@ class JobManagementResource @Inject()(val jobScheduler: JobScheduler,
       require(jobGraph.lookupVertex(jobName).isDefined, "Job '%s' not found".format(jobName))
       val job = jobGraph.getJobForName(jobName).get
       log.info("Manually triggering job:" + jobName)
-      jobScheduler.taskManager.enqueue(TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0), job.highPriority)
+      jobScheduler.taskManager.enqueue(TaskUtils.getTaskId(job, DateTime.now(DateTimeZone.UTC), 0, Option(arguments).filter(_.trim.nonEmpty))
```
jdef: I'm not sure how job.arguments is scrubbed or sanitized, but looking at this ... seems like someone could easily embed a semi-colon and then append some other command after it - getting two for the price of one. You know, a little shell script injection. Are we worried about those kinds of things here? Or can we assume that someone else already scrubbed (or will scrub )the `arguments` query parameter?

kensipe: its a very good point.   It seems that is missed in other parts of the code as well.   This was suppose to be just a bug fix.   That and the fact we are looking to move the whole thing to metronome I'm not sure if makes sense to spend too much time on it but I will have a look.  At least to get rid of the semi-colon.

jdef: If it's a PITA and exists in other parts of the codebase as well, maybe
just file a GH issue for it (because it's a problem in general) so as to
not block this PR

kensipe:
added protection against the evils of the world

kensipe:
@jdef thanks!

PR #732, opened 2016-11-14
Potential fix for mesos/chronos#553

### LINKED ISSUE #553: Colons in TaskID are problematic
From `src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala`:

```scala
//TaskIdFormat: ct:JOB_NAME:DUE:ATTEMPT:ARGUMENTS
val taskIdTemplate = "ct:%d:%d:%s:%s"
```

The colon `:` characters in that string are used by Mesos when constructing a temporary execution directory.

Unfortunately, putting colons in a Linux file path can be problematic. I attempted to execute a job that creates a Makefile then executes it. GNU Make reacts very poorly to either spaces or colons in build recipes.

For example, GNU Make doesn't handle this type of rule correctly:

```make
/foo/bar:fiz/buz: /some/other/file
    $(CC) -o $@ $^
```

Even though this is more of a problem with GNU Make than Chronos, I think it would be wise to use more normal delimiter characters such as `.` or `-` to separate out the parts of the task ID.

manossaratsis:
This patch fixes #553 by replacing ":" with "~" in the path. This addresses faulty behavior when using a chronos executor path in a variable like `$PATH`, `$PYTHONPATH`, `$CLASSPATH`, etc. which typically uses a ":" to separate paths.

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/TaskUtils.scala
@@ -18,10 +18,10 @@ import scala.collection.mutable
 
 object TaskUtils {
 
-  //TaskIdFormat: ct:JOB_NAME:DUE:ATTEMPT:ARGUMENTS
-  val taskIdTemplate = "ct:%d:%d:%s:%s"
+  //TaskIdFormat: ct~JOB_NAME~DUE~ATTEMPT~ARGUMENTS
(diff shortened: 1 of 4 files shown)


brndnmtthws:
I'd prefer not to change the task ID format, as it will break backward compatibility.

jerluc:
@brndnmtthws what suggestions would you have around using the Chronos executor path as part of the JVM classpath?

brndnmtthws:
You can use filesystem isolation (aka `chroot`) through docker or some other mechanism.

PR #733, opened 2016-11-16
support for user networks when using docker-containers

kaibra:
Hey,

when playing around with a docker-compose setup in which I am starting a mesos-cluster and a chronos instance, I saw, that chronos does not yet support custom user-networks as described in [http://mesos.apache.org/documentation/latest/networking/](url):

> For NONE, HOST and BRIDGE network mode the framework only needs to specify the network mode in the DockerInfo protobuf. However, for the USER mode, since a user-defined docker network is identified by a canonical network name (similar to CNI networks) apart from setting the network mode in DockerInfo the framework also needs to specify the name field in the NetworkInfo protobuf corresponding to the name of the user-defined docker network.

This pull-request would add the capability to specify USER-networks in the JSON which describes the job.
If you have a custom docker network running (e.g. created by docker-compose) you can connect to it by specifying this:

```json
{
  "container": {
    "type": "DOCKER",
    "image": "your-image",
    "network": "custom-network"
  }
}
``` 
which would create the following ContainerInfo in the MesosTaskBuilder:

```
type: DOCKER
docker {
  image: "your-image"
  network: USER
}
network_infos {
  name: "custom-network"
}
```

Let me know what you think.

Cheers,

Kai

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
@@ -9,14 +9,7 @@ object VolumeMode extends Enumeration {
   val RW, RO = Value
 }
 
-object NetworkMode extends Enumeration {
-  type NetworkMode = Value
 
-  // Bridged and Host
-  val BRIDGE, HOST = Value
-}
-
-import org.apache.mesos.chronos.scheduler.jobs.NetworkMode._
 import org.apache.mesos.chronos.scheduler.jobs.VolumeMode._
 
 case class Volume(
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
```diff
@@ -9,14 +9,7 @@ object VolumeMode extends Enumeration {
   val RW, RO = Value
 }
 
-object NetworkMode extends Enumeration {
```
brndnmtthws: Why'd you remove this? Why not just add the new enum value?

kaibra: Doing that, you would map a String to your custom Enum, which would then also have to store the user-network-name, which would then again be mapped to the `DockerInfo.Network.*` values in the MesosTaskBuilder.

What is the advantage of having the custom-enum?

If you remove the enum, the network just stays a String and the logic of mapping it to the Mesos API stays in the MesosTaskBuilder.

brndnmtthws: I'd prefer to keep the existing style the way it is.

brndnmtthws:
It looks like this is superseded by #763 now.

PR #756, opened 2016-12-07
Job name validation fix

xkrt:
Documentation declares that job name must `Must match the following regular expression: ([\w\s\.#_-]+)` but in fact Choronos backend uses other regexp.


xkrt:
I'm working on support of v3.0.0 in https://github.com/asher/chronos-python and find that old test doesnt passed https://github.com/asher/chronos-python/blob/master/itests/chronos-python.feature#L8

brndnmtthws:
Thanks for the PR, but this is intentional. I'd prefer to follow Marathon's convention and limit the allowed  characters (especially whitespace).

brndnmtthws:
I'll update the docs to reflect the change later.

Issue #757, opened 2016-12-08
Use sbt instead of maven


christobill:
What would be needed for a full build/release process with sbt instead of maven?
scoverage?
sbt-native-packager for docker building?

PR #758, opened 2016-12-08
#757 : sbt setup

christobill:


brndnmtthws:
We also modify the build version in the Docker build (https://github.com/mesos/chronos/blob/master/build-release.sh#L35). Do you know how to do this with sbt?

We'd need to update the build script to use sbt, too.

PR #750, opened 2016-12-03
Make sure we don't double-launch jobs.

brndnmtthws:

PR #762, opened 2016-12-13
Don't double enqueue tasks.

brndnmtthws:
A task which was already queued should not be re-enqueued.

This is to address an issue not fully address by #750.


andrew-t-l:
Yo, I work with @dave-handy, is there anyway you can add this as a tagged image in dockerhub, or should I build from scratch?

brndnmtthws:
The build is pending. It just takes a while. Just keep checking https://hub.docker.com/r/mesosphere/chronos/tags/

dave-handy:
Working so far! Thank you!

PR #708, opened 2016-07-28
WIP - Unified containers, cni networks and persistent/external volumes support.

vixns:
New features require mesos >= 1.0.0-rc1, should be compatible with mesos  >= 0.28 (not tested)
- Current limitations:
  - Lack APPC images support.
  - Lack image credential support.
  - WIP, need tests and documentation.

sample : 

``` json
{
  "schedule": "R/2016-01-01T00:00:00Z/PT1M",
  "name": "test",
  "container": {
    "type": "MESOS",
    "forcePullImage": true,
    "image": "debian",
    "networkName": "mynet",
    "volumes": [
      {
        "mode": "RW",
        "containerPath": "/tmp",
        "external": {
          "name": "test",
          "provider": "local-persist",
          "options": [
            {
              "key": "mountpoint",
              "value": "/tmp/test"
            }
          ]
        }
      },
      {
        "mode": "RW",
        "containerPath": "testp",
        "persistent": {
          "size": 100
        }
      }
    ]
  },
  "cpus": "0.1",
  "mem": "32",
  "command": "date"
}
```

PR #763, opened 2016-12-13
Unified containers, cni networks and external volume support

vixns:
see #708


sample.json
```diff
@@ -0,0 +1,36 @@
+{
```
brndnmtthws: What's this file for?

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
```diff
@@ -16,17 +16,37 @@ object NetworkMode extends Enumeration {
...
 case class Volume(
                    @JsonProperty hostPath: Option[String],
                    @JsonProperty containerPath: String,
-                   @JsonProperty mode: Option[VolumeMode])
+                   @JsonProperty mode: Option[VolumeMode],
+                   @JsonProperty persistent: Option[PersistentVolume],
+                   @JsonProperty external: Option[ExternalVolume])
 
-case class DockerContainer(
+case class Container(
```
brndnmtthws: 👍

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
```diff
@@ -16,17 +16,37 @@ object NetworkMode extends Enumeration {
...
                             @JsonProperty image: String,
+                            @JsonProperty `type`: ContainerType = ContainerType.DOCKER,
```
brndnmtthws: Why is `type` in backticks? It seems odd to me.

src/test/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilderSpec.scala
```diff
@@ -25,13 +25,13 @@ class MesosTaskBuilderSpec extends SpecificationWithJUnit with Mockito {
...
   val job = {
     val volumes = Seq(
-      Volume(Option("/host/dir"), "container/dir", Option(VolumeMode.RW)),
-      Volume(None, "container/dir", None)
+      Volume(Option("/host/dir"), "container/dir", Option(VolumeMode.RW), None, None),
+      Volume(None, "container/dir", None, None, None)
     )
 
-    var parameters = scala.collection.mutable.ListBuffer[Parameter]()
+    val parameters = scala.collection.mutable.ListBuffer[Parameter]()
```
brndnmtthws: 👍

brndnmtthws:
I suppose one use case might be that the user has some volume outside of the sandbox, which they want to use for storing some intermediate job data. However, we should call out in the docs that these volumes aren't managed in any way by Chronos. In other words, when the task ends, the data would be 'lost'.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
vixns: "type" is a scala reserved word

vixns:
You're right, we shouldn't add persistent local volumes without the concept of resident tasks.
I will remove them from this PR, I don't have time/interest to properly add resident tasks for now.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
```diff
@@ -174,16 +174,51 @@ class MesosTaskBuilder @Inject()(val conf: SchedulerConfiguration) {
...
+            .setDocker(Image.Docker.newBuilder()
+              .setName(job.container.image)
+              // TODO add setCredential
+              .build())
+            .setCached(!job.container.forcePullImage)
+            .build())
+          .build())
+    }
+    job.container.networkName.foreach {
+      n => builder.addNetworkInfos(NetworkInfo.newBuilder()
```
brndnmtthws: Maybe we should use the same spec that Marathon uses? As per https://mesosphere.github.io/marathon/docs/ip-per-task.html

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
```diff
@@ -16,17 +16,37 @@ object NetworkMode extends Enumeration {
   val BRIDGE, HOST = Value
```
brndnmtthws: Will need to add `USER` network type here to support user networks correctly.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
vixns: Agrees this is not sufficient, I keep this minimal implementation for now, because I (and maybe others) have many deployed jobs using this parameter, we'll add a more complete networkinfo support in the future.

`name` is the only required param for CNI networks, other params, like `protocol`, `labels` or `port_mappings` are optional.

The marathon's ip-per-task spec does not support multiple networks.
IMO, we should use a spec closer to the networkInfo proto https://github.com/apache/mesos/blob/81cd023eb9945a22c220edc966393dcfcdbce256/include/mesos/mesos.proto#L1922

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
vixns: Ok, I'll add USER, even if it's useless for unified containers.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
brndnmtthws: I'm hesitant to accept this if it's incomplete, or if there's risk of API change. I'd prefer to have the API right from the start so we don't get 'stuck'.

Agreed on the last comment. I guess we should do something like this:

```json
{
  "container":{
    "type":"MESOS",
    "forcePullImage":true,
    "image":"debian",
    "networkInfos":{
      "name":"mynet"
    }
  }
}
```

vixns: What about keeping `networkName` for simple cases, and add 
```json
"networkInfos": [
{
  "name": "mynet",
  "protocol": "IPv4",
  "labels": [
    {
      "key": "mylabel",
      "value": "myvalue"
    }
  ],
  "port_mappings": [
    {
      "host_port": 80,
      "container_port": 8080,
      "protocol": "tcp"
    }
  ]
}
]
```
There are no side effects if both are present, a `"networkName":"mynet"` is the same as

```json
"networkInfos": [
{
  "name": "mynet"
}
]
```

brndnmtthws: Okay, fair enough.

👍

docs/docs/api.md
```diff
@@ -278,6 +281,85 @@ There is also support for passing in arbitrary docker config options.
...
 
+## Adding a Containerized Job
+
+A containerized job takes the same format as a scheduled job or a dependency job and runs on a Mesos container.
+To configure it, an additional `container` argument is required, which contains a type (required), an image (required), a network name (optional), mounted volumes (optional) and whether Mesos should always pull the latest image before executing or not (optional).
+
+```json
+{
+  "container": {
+    "type": "MESOS",
```
brndnmtthws: The UI should probably be updated at some point to support this as well.

src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala
brndnmtthws: Should also add `USER` here: https://github.com/mesos/chronos/blob/master/src/main/resources/ui/models/JobForm.js#L217-L221

I just noticed a copy-paste artefact after that `return` statement too.

brndnmtthws:
@vixns just wondering if you have time to work on this this week. If not, I'll pick up where you left off as I'd like to get it merged and push out a v3.0.1 release.

vixns:
I'd like to see this integrated in 3.0.0, I'll try to finish this before friday.

PR #766, opened 2016-12-14
Fetcher should extract by default, if not defined.

brndnmtthws:


brndnmtthws:
cc @turb

turb:
@brndnmtthws since it is Mesos default, I agree we should follow it.

cf. [Mesos fetcher documentation](https://mesos.apache.org/documentation/latest/fetcher/)

PR #770, opened 2016-12-29
Adding support to mesos-role

vidhyaarvind:
Adding support to single role. Default is still * role

### DIFF
--- a/src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala
+++ b/src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala
@@ -73,7 +73,7 @@ class MesosJobFramework @Inject()(
 
   def getReservedResources(offer: Offer): (Double, Double) = {
     val resources = offer.getResourcesList.asScala
-    val reservedResources = resources.filter({ x => x.hasRole && x.getRole != "*" })
+    val reservedResources = resources.filter({ x => x.hasRole && x.getRole != config.mesosRole()})
     (
       getScalarValueOrElse(reservedResources.find(x => x.getName == "cpus"), 0),
       getScalarValueOrElse(reservedResources.find(x => x.getName == "mem"), 0)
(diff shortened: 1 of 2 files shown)


vidhyaarvind:
@brndnmtthws could you please review this?

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala
```diff
@@ -73,7 +73,7 @@ class MesosJobFramework @Inject()(
 
   def getReservedResources(offer: Offer): (Double, Double) = {
     val resources = offer.getResourcesList.asScala
-    val reservedResources = resources.filter({ x => x.hasRole && x.getRole != "*" })
+    val reservedResources = resources.filter({ x => x.hasRole && x.getRole != config.mesosRole()})
```
brndnmtthws: This is actually the reverse logic. The original version of this method was correct.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
```diff
@@ -36,7 +36,7 @@ class MesosTaskBuilder @Inject()(val conf: SchedulerConfiguration) {
       .setName(name)
       .setType(Value.Type.SCALAR)
       .setScalar(Value.Scalar.newBuilder.setValue(value))
-      .setRole("*")
+      .setRole(conf.mesosRole())
```
brndnmtthws: This doesn't look right either.

vidhyaarvind: @brndnmtthws Resource was taking * as the hardcoded value. Now the conf value set in  --mesos-role is used. I have tested in mesos using multiple chronos frameworks and various roles. My cluster now does not have * roles. Also using Resource.newBuilder we can only setRole. Its's not possible to add multiple roles.

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala
vidhyaarvind: I can double check to see if this rejects * role offers, my cluster currently does not have any agents with * role

PR #804, opened 2017-02-16
Add missing MESOS_TASK_ID environment variable

lionelnicolas:
As `MESOS_TASK_ID` is an environment variable, it should be uppercase.


lionelnicolas:
Ok I've updated the PR to keep the lowercase variable as well.

Issue #764, opened 2016-12-14
Chronos + GPU resources


routehero:
Mesos supports GPU resources now but there doesn't appear to be a way to schedule a job that requires a GPU resource.

PR #810, opened 2017-02-26
#764 Add GPU support via --enable-features

reneploetz:
This patch adds GPU scheduling support to Chronos the same way Marathon does it:
- you need to specify "--enable-features=gpu_resources" as parameter to enable it
- you can specify "--mesos_task_gpu=X" where X is an Integer to always allocate X GPUs for new tasks
- the REST API was extended to allow an optional field "gpus" alongside the usual "cpus/disk/mem" arguments to allow per-task GPU settings

The argument parsing code was adapted from Marathon (Features.scala / SchedulerConfiguration.scala) to provide the same behaviour.

Signed-off-by: Rene Ploetz <reneploetz@gmx.de>


src/test/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilderSpec.scala
```diff
@@ -36,7 +36,7 @@ class MesosTaskBuilderSpec extends SpecificationWithJUnit with Mockito {
 
     val parameters = scala.collection.mutable.ListBuffer[Parameter]()
 
-    val container = Container("dockerImage", ContainerType.DOCKER, volumes, parameters, NetworkMode.HOST, None, networks, forcePullImage = true)
```
brndnmtthws: Please add a new test instead of changing this one.

reneploetz: I'll change the type back to DOCKER.
Do you want some tests for the MesosTaskBuilderSpec, e.g. whether a non-zero value would be applied? And do you want the commits squashed?

gengmao: @reneploetz According to http://mesos.apache.org/documentation/latest/gpu-support/, seems DOCKER type is not the right one
>  At the time of this writing, Nvidia GPU support is only available for tasks launched through the Mesos containerizer (i.e. no support exists for launching GPU capable tasks through the Docker containerizer)

brndnmtthws:
@reneploetz let me clarify: I was suggesting you add an additional (new) test with the GPUs enabled, and leave the existing test unchanged.

Issue #816, opened 2017-03-16
WTB: Confirmation before deleting job on web ui


thekad:
In Chronos 3.0.2, clicking the red X button a job's actions will immediately remove the job. It would be awfully nice if it waited for a confirmation that you indeed want to remove said job.

thekad:
:man_shrugging:

PR #820, opened 2017-03-29
Ask for confirmation before taking action in job summary

thekad:
This is related to #816, putting a confirmation dialog before running/stopping/deleting jobs seems like a good idea. Not using any fancy react libraries just good ol' javascript.
I am not specially sure about 7cfe1f3900 but it _seems_ to work just fine, I believe it to be a typo because the function definition for ``deleteJob()`` in ln181 expects an event.


thekad:
The travis failure seems unrelated to my change, as far as I can see?

thekad:
Seems this repo has been abandoned.

PR #822, opened 2017-03-30
only record task launch with the offer used

meelapshah:
This typo caused chronos to think that each task was being launched with each offer.


meelapshah:
I guess not with each offer, but rather identical (task, slave) launchings were recorded, once for each offer in the offer set.

PR #823, opened 2017-04-06
Configurable notification level

andresilva:
This PR adds basic notification level support. Three notification levels are provided: `DISABLED`, `FAILURES`, `ALL`. Currently the following notifications will be affected:

* `JobRemoved` - only sent if notification level is set to `ALL`
* `JobDisabled` - only sent if notification level is set to `ALL`
* `JobRetriesExhausted` - sent if notification level is set to `FAILURES` or `ALL`

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -51,6 +51,8 @@ If you installed chronos via package, run `/usr/bin/chronos run_jar --help`.
                                             otherChronos instances and the
                                             Mesos master to communicate with
                                             this instance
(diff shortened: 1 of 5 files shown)


andrebeat:
This is currently very much bare-bones since I've only implemented the minimum necessary for my use case but I believe this in a state that is easily extendable (e.g. adding more notification levels). Let me know what you think.

andrebeat:
Sorry to bump this but are PRs still being reviewed/merged?

PR #831, opened 2017-05-10
Bugfix prop-types

thekad:
React 15.5 split off ``PropTypes`` to its own package ``prop-types``. The way ``package.json`` specifies versions made it to download this new version and update several other dependencies.
This is the least intrusive change I could come up with, fixing the version to 15.4.x and also pinning `react-input-autosize` because of JedWatson/react-input-autosize#84


thekad:
This project has been abandoned it seems.

PR #835, opened 2017-06-16
[Bug Fix] Ace editor cursor position, caused by non-monospaced font

Vashnak:
In some browsers/versions (I do not know which ones) we could see a bug during json edition. Cursor war mispositionned because the font used was not monospaced. Added a css rule to resolve it.


Vashnak:
test failure is unrelated to my change since I only updated a css file

PR #854, opened 2017-08-29
Don't kill executors sending a TASK_STARTING update.

lava:
Since the dawn of Mesos, executors have had the option of sending a `TASK_STARTING` update to indicate that the launch command has been received. Currently, Chronos will react to to this update by killing the task.

As far as I can tell, this should fix the issue and tests are passing, but I will admit that I have no experience with either Scala or Maven.


gkleiman:
LGTM

How did you test this? Did you try setting up a cluster with a version of Mesos that sends `TASK_STARTING` and checking if Chronos works fine?

Can you please open a similar PR against the `stable` branch?

Thanks!

gkleiman:
I synced with @lava, he did some manual testing and the change works fine.

PR #884, opened 2018-09-06
Add max completion time to taskInfo (#881)

edi33416:


edi33416:
This cherry-picks #881 onto stable, in order to provide `maxCompletionTime` to version 2.x

@gkleiman can you please review?

edi33416:
Please note that I had to remove the `concurrent` field, as it wasn't present on `stable`.
See errors
```
[ERROR] /home/fawkes/chronos/src/main/scala/org/apache/mesos/chronos/utils/JobSerializer.scala:130: error: value concurrent is not a member of org.apache.mesos.chronos.scheduler.jobs.BaseJob
[ERROR]     json.writeBoolean(baseJob.concurrent)
```

```
[ERROR] /home/fawkes/chronos/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala:127: error: value concurrent overrides nothing
[ERROR]                              @JsonProperty override val concurrent: Boolean = false,
[ERROR]                                                         ^                                                                                                                                                  
[ERROR] /home/fawkes/chronos/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Jobs.scala:166: error: value concurrent overrides nothing
[ERROR]                                @JsonProperty override val concurrent: Boolean = false,
[ERROR]                                                           ^  
```

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
```diff
@@ -127,6 +128,13 @@ class MesosTaskBuilder @Inject()(val conf: SchedulerConfiguration) {
       }
     }
 
+    // Add maxCompletionTime
+    if (job.maxCompletionTime > 0L) {
+      val maxCompletionTime = DurationInfo.newBuilder().setNanoseconds(job.maxCompletionTime * pow(10, 9).toLong).build()
```
gkleiman: I should have noticed this on the previous PR, but better late than never =).

I think we should use `scala.concurrent.duration.Duration` here, like this:

```
val maxCompletionTime = DurationInfo.newBuilder().setNanoseconds(Duration(job.maxCompletionTime.seconds.toNanos).build()
```

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
```diff
@@ -127,6 +128,13 @@ class MesosTaskBuilder @Inject()(val conf: SchedulerConfiguration) {
...
+      taskInfo.setMaxCompletionTime(maxCompletionTime)
+      taskInfo.setMaxCompletionTime(maxCompletionTime)
```
gkleiman: Remove this line.

gkleiman (changes requested):
Thanks! Just two minor comments =).

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
edi33416: Oops. Will remove

src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala
```diff
@@ -127,6 +129,14 @@ class MesosTaskBuilder @Inject()(val conf: SchedulerConfiguration) {
       }
     }
 
+    // Add maxCompletionTime
+    if (job.maxCompletionTime > 0L) {
+      val maxCompletionTime = DurationInfo.newBuilder()
+            .setNanoseconds(job.maxCompletionTime.seconds.toNanos)
```
edi33416, line 135: @gkleiman is this ok, or would you prefer
`.setNanoseconds(Duration(job.maxCompletionTime, SECONDS).toNanos)` ?

gkleiman:
LGTM, thank you so much!

